PR #46, opened 2014-12-31
Add getters for docParser/HtmlRenderer/extensions

dshafik:
These small additions make it much easier to make changes to the parser/renderer/environment (and helps to be able to effectively unit test them a bit better)

**Added**: made *all* `private` elements `protected` instead, kept running into issues when extending. A concrete example is that I am unable to easily add support for `?[Word](definition)` as most of `League\CommonMark\Inline\Parser\CloseBracketParser` is private.

### DIFF
--- a/src/CommonMarkConverter.php
+++ b/src/CommonMarkConverter.php
@@ -58,4 +58,20 @@ class CommonMarkConverter
 
         return $this->htmlRenderer->renderBlock($documentAST);
     }
+
+    /**
+     * @return DocParser
+     */
+    public function getDocParser()
+    {
+        return $this->docParser;
+    }
+
+    /**
+     * @return HtmlRenderer
+     */
+    public function getHtmlRenderer()
+    {
+        return $this->htmlRenderer;
+    }
 }
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -25,6 +25,11 @@ use League\CommonMark\Inline\Renderer\InlineRendererInterface;
 class Environment
 {
     /**
+     * @var ExtensionInterface[]
+     */
+    protected $extensions;
+    
+    /**
      * @var BlockParserInterface[]
      */
     protected $blockParsers = array();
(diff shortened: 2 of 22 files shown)


src/CommonMarkConverter.php
```diff
@@ -58,4 +58,20 @@ public function convertToHtml($commonMark)
...
+        return $this->docParser;
+    }
+
+    /**
+     * @return HtmlRenderer
+     */
+    public function getHtmlRenderer()
+    {
+        return $this->htmlRenderer;
+    }
```
colinodell: > when all I want is to change one thing in the htmlRenderer

There's nothing you can really change though - if there was, then I'd be okay with this change.

src/Environment.php
```diff
@@ -245,6 +252,14 @@ public function addExtension(ExtensionInterface $extension)
...
 
     /**
+     * @return ExtensionInterface[]
+     */
+    public function getExtensions()
+    {
+        return $this->extensions;
+    }
+
+    /**
```
colinodell: Okay that makes sense.

src/HtmlRenderer.php
```diff
@@ -142,4 +142,12 @@ public function renderBlocks($blocks, $inTightList = false)
...
         return implode($this->options['blockSeparator'], $result);
     }
+
+    /**
+     * @return Environment
+     */
+    public function getEnvironment()
+    {
+        return $this->environment;
+    }
```
colinodell: I'm just worried about adding and supporting new public methods  solely for testing purposes (no strong use-case otherwise).

colinodell:
I'm closing this in favor of #52 and #53:

 - #52 includes one of the getters you were looking for.  I'd prefer to avoid adding the other unless there's a compelling use-case for them.
 - #53 includes the majority of the private -> protected changes.  I kept a few things private that I didn't think were necessary to expose.  If you disagree, let me know which ones you think should change and why.

Thanks!

PR #48, opened 2015-01-02
Allow for multiple reference parsers

dshafik:
This extends PR #46 but should be possible to separate

- Makes it possible to register ReferenceParser's (plural) from the extension
- Does NOT currently append ReferenceParsers, this is different to Parsers/Renderers/InlineProcessors
- I think this behavior should be changed for those, as you can use OO principles to extend syntaxes, allowing for any ordering gymnastics to take place. Also, the functions are `set*` not `add*` (except in the case of `addExtension`)

### DIFF
--- a/src/Block/Element/Document.php
+++ b/src/Block/Element/Document.php
@@ -23,7 +23,7 @@ class Document extends AbstractBlock
     /***
      * @var ReferenceMap
      */
-    private $referenceMap;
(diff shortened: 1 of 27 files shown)


colinodell:
I'm going to close this because I don't feel we've identified any compelling use-cases which can't be solved some other (better) way.

PR #49, opened 2015-01-02
UTF-8 support

colinodell:
Spec version 0.14 enhanced some of the rules regarding emphasis parsing.  It's now necessary to [peek around at characters in a Unicode-aware way](http://spec.commonmark.org/0.14/#right-facing-delimiter-run).

This introduces a major problem - the `Cursor` class is not currently Unicode-aware.  For example, it thinks `[Толпой]` is 14 characters long instead of 8.  This was fine for our purposes since we never needed to address Unicode characters by position.

So the new challenge is that given a `Cursor` in some position, we need to accurately obtain neighboring Unicode characters.  Our `$this->line[$this->position]` trick won't work as it will only obtain a single byte instead of the whole character.  The proper solution would require the `Cursor` becoming Unicode-aware.

`iconv` and `mbstring` both seemed to be strong candidates for this task, so I implemented both and benchmarked them.  `iconv` resulted in a 1000% performance penalty, whereas `mbstring` represented a 26% drop, so I implemented the latter where needed.

As much as I hate adding new dependencies and reducing performance, I do feel this is the correct approach.  I definitely welcome any feedback, alternatives, or performance enhancements.  I'll keep this open for a few days to gather feedback before accepting.

/cc @philsturgeon @cebe @GrahamCampbell @aleemb @dshafik


GrahamCampbell:
We could still use the patchwork library you know.

colinodell:
I looked into [patchwork/utf8](https://github.com/tchwork/utf8) as one possibility, but it seems to fallback to `iconv` which is terrible for performance.

GrahamCampbell:
Requiring the extension is probably the best way to go then. :)

mnapoli:
Do you guys know if mbstring is that widespread? I have no idea

colinodell:
I tried researching that but wasn't able to find any conclusive information.  It does seem to be available in the most-popular Linux distros though:

* Debian/Ubuntu: built-in by default
* RHEL/CentOS and Fedora: `yum` package available
* openSUSE: `yast` package available

mnapoli:
Thanks!

PR #55, opened 2015-01-07
Bugfix for HTML Element __toString()

tjlytle:
Logic was too broad, any value that could be considered false (like '0') would be treated as an empty tag.

### DIFF
--- a/src/HtmlElement.php
+++ b/src/HtmlElement.php
@@ -122,7 +122,7 @@ class HtmlElement
             $result .= ' ' . $key . '="' . $value . '"';
         }
 
-        if ($this->contents) {
+        if (!is_null($this->contents) AND $this->contents != '') {
             $result .= '>' . $this->getContents(true) . '</' . $this->tagName . '>';
         } elseif ($this->selfClosing) {
             $result .= ' />';
--- a/src/HtmlElement.php
+++ b/src/HtmlElement.php
@@ -122,7 +122,7 @@ class HtmlElement
             $result .= ' ' . $key . '="' . $value . '"';
         }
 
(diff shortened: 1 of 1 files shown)


src/HtmlElement.php
```diff
@@ -122,7 +122,7 @@ public function __toString()
             $result .= ' ' . $key . '="' . $value . '"';
         }
 
-        if ($this->contents) {
+        if (!is_null($this->contents) and $this->contents != '') {
```
colinodell: `$this->contents` should never be `null`, so we could remove that first check.

philsturgeon: It should also probably be `!== ''`

GrahamCampbell: also, we should use `&&` over `and`.

philsturgeon: Thanks Graham.

@colinodell with the suggested change made, tests are reporting the following: 

```
1) League\CommonMark\Tests\HtmlElementTest::testToString
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'<p></p><img />'
+'<p></p><img></img>'
```

tjlytle:
Yeah, just looked at property declaration and the constructor, didn't think far enough that it shouldn't ever be `null`.

src/HtmlElement.php
colinodell: Ugh I'm probably using `null` somewhere I shouldn't be.  /facepalm

I'll look into this.

colinodell: Yep, that was my bad: https://github.com/thephpleague/commonmark/commit/34884ce26b1a571d46c5d8ff7c6bc172e45aea13

Issue #42, opened 2014-12-30
Configuration


colinodell:
Parsers and renderers will need to be configurable at some point.  For example, some users may want to whitelist certain HTML elements as "safe", ensuring that others get escaped.  If we were to allow such configurations now, the user would need to manually build their `Environment` and set the options on each individual parser/renderer:

~~~php
$environment = new Environment();
$linkParser = new LinkParser(['defaultScheme' => 'https']);
$htmlParser = new HtmlParser(['allowedTags' => ['a', 'strong', 'p', 'dl', 'dt', 'dd']]);
// ... Repeat for every single parser/renderer

// Also repeat this for every new parser/renderer you add
$environment->addInlineParser($linkParser);
// ... etc
~~~

It would be much easier if the user could obtain the pre-built `Environment` and simply pass all their configurations in at once.  Perhaps it would look like this:

~~~php
$environment = Environment::createCommonMarkEnvironment([
    'links' => [
        'defaultScheme' => 'https',
    ],
    'html' => [
        'allowedTags' => ['a', 'strong', 'dl', /*...*/ ]
    ],
]);
$docParser = new DocParser($environment);
$htmlRenderer = new HtmlRenderer($environment);
~~~

So the questions I have are:

1. Is this a good idea?
2. How do the parsers/renderers define which options they support?
3. How do the config values get set on the parsers/renderers?
4. Do we use something like symfony/options-resolver, simple arrays, or roll our own solution?
5. Should users be able to change the configuration on-the-fly, or is it an immutable private/protected variable than cannot be changed once the Environment is created?

colinodell:
Thanks for the feedback!  After considering this for the past week, I've come to these conclusions:

1. Configuration options are necessary.
2. Parsers/renders won't specify their required options, as this is more complex than it really needs to be right now.  This can be revisited later if needed.
3. Options will be set as a simple array on the `Environment`.  Parsers/renders will be able to access the full array to look at whatever options they need.
4. We'll use a simply config array.  Parsers/renders can use nested array merges to combine the given config with their defaults.
5. The config would ideally become "locked" once extensions have been initialized, though I may not enforce this in the next release.

colinodell:
I have implemented a possible solution in PR #56.  I'd greatly appreciate any feedback.

PR #56, opened 2015-01-08
Configuration

colinodell:
See #42


colinodell:
My only question is whether the temporary deprecation stuff is really necessary, or if we can just strip that out and use the new methodology only.

colinodell:
I've decided I'd like to strip [the deprecated warning stuff](https://github.com/thephpleague/commonmark/commit/d937a6d0563cbdd5dc5e1243a60e2cde1c3ef641#diff-114d0439c76db6068924dcfc2ad0997eR44).  It's technically unnecessary since we're jumping to 0.6 and it represents extra (untested!) functionality to support.  We'll just be super clear about these changes in the changelog.

PR #67, opened 2015-02-16
Bumped branch alias

GrahamCampbell:
I noticed a public method was removed, among other things since 0.6.1.


colinodell:
Good catch, thanks!

PR #69, opened 2015-02-16
Optimizations

colinodell:
We now parse long spans of non-special characters at once, vs the old approach which parsed each character individually.  This results in a whopping **50% performance boost** according to our internal benchmark (310ms to 156ms on my machine).


GrahamCampbell:
Nice. :)

PR #72, opened 2015-02-18
Hopefully resolving some of the Scrutinizer alerts

rawkode:


colinodell:
Thanks for submitting this!

Scrutinizer failed to run for some reason so I've restarted it.  I'll take a closer look once it completes its analysis.

src/DocParser.php
```diff
@@ -98,7 +98,7 @@ private function incorporateLine(ContextInterface $context)
         $context->setContainer($context->getDocument());
         while ($context->getContainer()->hasChildren()) {
             $lastChild = $context->getContainer()->getLastChild();
-            if (!$lastChild->isOpen()) {
+            if (!is_null($lastChild) && !$lastChild->isOpen()) {
```
colinodell: I don't think this check is necessary.  Line 99 checks whether any children exist, so line 100 will always return a child object.

rawkode: Ah, you're correct! I'll remove

rawkode:
Scrutinizer seems to be complaining with

```Error while resolving doc type "$regex": Unknown type name "$regex" at position 0.```

colinodell:
I restarted Scrutinizer again (I think it timed out waiting on Travis).  Anyway, these changes look good.  Thanks so much!

PR #76, opened 2015-02-25
Drop support for PHP 5.3

colinodell:


colinodell:
Support for 5.3 will be dropped by the end of the week.  If you feel it's too early for this change, please let me know why, otherwise I shall proceed with this merge.

cebe:
What's the point in dropping support for a version without any technical reason? I would understand if you'd want to use traits now which are not supported in 5.3...

GrahamCampbell:
php 5.3 is eol, so anyone using it now is using it at risk of being exploited. There are known, unfixed security vulnerabilities.

cebe:
I know that and I agree that one should not use it anymore and upgrade as soon as possible.
But imo that is not the job of a composer.json requirement, this has to be done on a different level.
The requirements in composer.json should reflect the technical requirements of a library and not what the author thinks is an appropriate php version to be installed on users system.

It is good to point out that 5.3 is old and should not be used but patronising people that may be stuck with 5.3 for some reason and may be in the progress of upgrading is an unnecessary hurdle.

colinodell:
We had a good discussion about this on the League's mailing list: https://groups.google.com/d/msg/thephpleague/632Kzor2hu8/rwJXs-wLWc8J

There have been at least three instances where supporting 5.3 prevented us from implementing positive changes:

* A user-contributed PR which used traits to simplify the codebase (5.4)
* The old block-closing closure couldn't reference `$this`, so we had to set `$self = $this;` and then `use ($self)` to workaround that limitation (5.4)
* The `Environment` configuration references element types using their full class name.  Would be nice if we could use the special `::class` constant instead of hard-coding the full namespaced class name (5.5)

So while you're right that there's no immediate technical need, I do feel we are unnecessarily restricting ourselves from positive enhancements by supporting an unsupported version.

Anyone who can't upgrade from 5.3 is welcome to use the current 0.7 release which is fairly stable and feature-packed, so it's not like we're completely abandoning them.

philsturgeon:
+1 to that last message. 

It's not like you're doing it for no reason, or purely for short arrays, but you're doing it to make larger useful changes which will benefit code cleanliness and performance.

cebe:
I am totally fine with doing it for technical reasons. It was not obvious from the changes in the PR and I did not know about the discussion thread. Go ahead and merge! :)

philsturgeon:
Yeah right there with you on that. Gotta be responsible with these choices https://philsturgeon.uk/blog/2013/03/pick-php-requirements-for-packages-responsibly/

cebe:
(y) https://github.com/cebe/markdown/issues/8#issuecomment-37549375

colinodell:
Just encountered a 5.3 bug: https://github.com/thephpleague/commonmark/commit/57478220078c1ffb66eef2777f554fbd57d6c307

Unfortunately this workaround is only needed for versions < 5.4.8 but it results in a ~5% performance hit to **all** versions :-/

My current plan is to address all remaining issues and release a final 0.7.x version.  After that I'll start working towards 0.8, which will drop 5.3 support and undo that workaround.

GrahamCampbell:
:+1: for only requiring newer versions of 5.4 and upwards. :)

PR #77, opened 2015-02-25
Test against PHP 7.0 (nightly) but allow failures

colinodell:
We'll remove `allow_failures` once Travis includes the `mbstring` extension and PHP 7.0 enters a relatively-stable state (the tests work reliably well)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,12 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
+
+matrix:
+  allow_failures:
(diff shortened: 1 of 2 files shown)


colinodell:
Unfortunately it seems that the nightly builds used by Travis do not include the `mbstring` extension, resulting in failures.  I will merge this anyway in anticipation of extensions being included in the PHP 7 builds.

philsturgeon:
Yeah just keep an eye on progress. I’ll be monitoring stuff so if I notice it passing I’ll poke you.

PR #87, opened 2015-04-13
Fix name of method argument

bcremer:
Seems like a typo or refactoring relict.

This should not change any public behavior or breake public APIs.


colinodell:
Thanks for catching this!  I'll get this merged once I get the HHVM tests to stop failing (unrelated).

PR #88, opened 2015-04-14
Minor phpdoc and PSR-2 fixes

bcremer:
While digging deeper in the code my IDE noticed some minor phpdoc and PSR-2 issues.

### DIFF
--- a/src/Block/Renderer/DocumentRenderer.php
+++ b/src/Block/Renderer/DocumentRenderer.php
@@ -21,9 +21,9 @@ use League\CommonMark\HtmlRendererInterface;
 class DocumentRenderer implements BlockRendererInterface
 {
     /**
-     * @param Document $block
+     * @param AbstractBlock|Document $block
+     * @param HtmlRendererInterface $htmlRenderer
      * @param bool $inTightList
-     *
      * @return string
      */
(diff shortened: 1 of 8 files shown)


src/Block/Renderer/DocumentRenderer.php
```diff
@@ -21,9 +21,9 @@
 class DocumentRenderer implements BlockRendererInterface
 {
     /**
-     * @param Document $block
+     * @param AbstractBlock|Document $block
+     * @param HtmlRendererInterface $htmlRenderer
      * @param bool $inTightList
-     *
```
colinodell: Could we keep this empty line between the `param` and `return`?  I like using the PEAR standard of having exactly one empty line in between.

colinodell: Actually nvm, I shouldn't be asking you to make more changes simply to please my inner OCD, especially when your improvements are great as-is.

colinodell:
Thanks for the contribution @bcremer!

src/Block/Renderer/DocumentRenderer.php
bcremer: Thanks for the merge. I can totally relate to your inner OCD in regards to coding styles ;)

Issue #89, opened 2015-04-14
Remove hard-coded phpunit-mock-objects version requirements


colinodell:
We're currently requiring version 2.3.0 because [2.3.1 uses a broken reflection implementation in HHVM](https://github.com/sebastianbergmann/phpunit-mock-objects/issues/223) which causes our tests to fail.  Once HHVM fixes [this issue](https://github.com/facebook/hhvm/issues/5170) I'd like to revert commit 2a6a97c so we're using the latest and greatest version again.

colinodell:
Huh, looks like Travis isn't using the new HHVM build yet.  Will put this back on hold until HHVM 3.5.0 is upgraded.

colinodell:
It looks like the fix landed in HHVM 3.8.0 but Travis is still using 3.5.0.  https://github.com/facebook/hhvm/commit/b786320582e086630bbeb65043c3be5680257ba0

PR #91, opened 2015-04-28
Test improvements

colinodell:
We may eventually add additional functional tests, so it would be nice to have a
separate area for those.


composer.json
```diff
@@ -25,7 +25,8 @@
         "jgm/CommonMark": "0.18",
         "michelf/php-markdown": "~1.4",
         "phpunit/phpunit": "~4.3",
-        "phpunit/phpunit-mock-objects": "2.3.0"
+        "phpunit/phpunit-mock-objects": "2.3.0",
```
GrahamCampbell: any reason why this is locked that that version?

colinodell: Yeah, 2.3.1 uses a reflection feature that's broken in HHVM.  I've temporarily locked this version until HHVM can resolve the issue:  https://github.com/thephpleague/commonmark/issues/89

PR #95, opened 2015-05-13
Added storage to block element for arbitrary data

hason:
Similarly as in inline element https://github.com/thephpleague/commonmark/blob/master/src/Inline/Element/AbstractInline.php#L24


colinodell:
Thanks for the contribution!

PR #96, opened 2015-05-16
Fixed phpdoc

GrahamCampbell:
The "short description" MUST be the first non-empty line in the phpdoc.


colinodell:
Good catch, thanks!

Issue #35, opened 2014-12-25
Support for multiple renderers


aleemb:
My use case is that I need custom renderers for images and depending on the image URL, these renders will render the image as an iframe for YouTube, Vimeo, etc. At present I use the Sundown renderer and I have stuffed all these renderers into one big file but would like to move over to this lib and cleanup the mess while I'm at it.

I also imagine that a lot of contributors can then share their own YoutubeRenderer (or generic VideoRenderer). This would make extensibility a lot more fun and build an ecosystem around it. Maybe even give way to an extensions repo.

Based on your suggestion in #32, I currently render my custom Renderer as

```
$environment->addInlineRenderer('League\CommonMark\Inline\Element\Image', new MediaRenderer());
```

However, this only allows attaching a single renderer. It would be nice if the renderer can work sort of like conventional event handlers with the possibility to either chain the output as input to other renderers or stop propagation. The easiest approach would be to have symmetric input/output for renderers that would make chaining a breeze. Currently the input/output is `(AbstractInline, HtmlRenderer)/HtmlElement`.

indrora:
Taking on my $0.002.

I'm attempting to implement a new kind of link type (internal references) and I'm having a hard time. I'm trying to make it such that ```^[some page](some/where/inside)``` can be parsed into what would be ```[some page](/some/path/some/where/inside)```

Having a way to mangle links like this would be *fantastic* -- it would give the ability to have what OP wants (Totally override the link process) but also fit several other use cases (e.g. a custom one to make your internal documentation, say `$[this CVE](2008-4250)` could become the equivalent of `[this CVE](http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2008-4250)`

bfrohs:
I implemented YouTube video rendering using a processor that would replace any `Link` that has a YouTube video URL with a `YouTubeVideo` element that extends `Link`, leaving the url, label, and link alone as they're passed through. Then I added a `YouTubeVideoRenderer` that ensures the link is by itself in a paragraph, pulls the video ID out of the URL, and creates the necessary elements to display the video.

This way, if the YouTube processor/render are ever removed, it will fallback to a link to the YouTube video in it's own paragraph, which doesn't look half bad and still serves the same purpose.

See https://gist.github.com/bfrohs/9017f38586404f161298 for code.

skyzyx:
[...]

Issue #32, opened 2014-12-23
Unable to add most attributes to images


aleemb:
I am writing a custom image processor that needs to add custom classes to an image:

```
$inline->attributes['class'] = 'foo';
```

However, the ImageRenderer explicitly sets the `alt`, `src` and `title` attributes only. Is this by design or can this be allow to pass through any other attributes?

Alternatively, I tried adding a custom renderer to add more attributes but am probably missing some pieces of the puzzle to get this working:

```
class MediaRenderer implements InlineRendererInterface
{
    public function render(AbstractBaseInline $inline, HtmlRenderer $htmlRenderer)
    {
        return $htmlRenderer->inTags('img', ['class'=>'foo'], '', true);
    }
}

$environment->addInlineRenderer('MediaRenderer', new MediaRenderer());
```

I then noticed that the ImageRender is added by default and there is no `removeInlineRenderer` option to unset it. Even so, creating my own image renderer for attribute support would require me to keep copy/paste the code from `ImageRenderer` and ensure it stays in sync in future releases.

Allowing other attributes besides `alt`, `src` and `title` would be the easiest so hoping that can come through :)

PR #101, opened 2015-05-27
Added support for tag attributes in renderers

hason:
I would like to implement special attributes extension and this PR would help me greatly.

### DIFF
--- a/src/Block/Element/AbstractBlock.php
+++ b/src/Block/Element/AbstractBlock.php
@@ -317,4 +317,9 @@ abstract class AbstractBlock
 
         return $this;
     }
+
+    public function getData($name, $default = null)
+    {
+        return array_key_exists($name, $this->data) ? $this->data[$name] : $default;
+    }
 }
--- a/src/Block/Renderer/BlockQuoteRenderer.php
+++ b/src/Block/Renderer/BlockQuoteRenderer.php
@@ -34,14 +34,19 @@ class BlockQuoteRenderer implements BlockRendererInterface
             throw new \InvalidArgumentException('Incompatible block type: ' . get_class($block));
         }
 
+        $attrs = array();
+        foreach ((array) $block->getData('attributes', array()) as $key => $value) {
+            $attrs[$key] = $htmlRenderer->escape($value, true);
+        }
+
         $filling = $htmlRenderer->renderBlocks($block->getChildren());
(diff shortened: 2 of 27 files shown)


colinodell:
Thanks for submitting this!

The ability to define arbitrary HTML attributes [has come up before](https://github.com/thephpleague/commonmark/issues/32#issuecomment-68006068), and your solution almost exactly matches my second idea :)  I still have the same reservations about using this approach though.

I wonder if we should modify rendering to be more event-based, or perhaps just add a generic post-render event.  You'd have some custom code listening for an event.  You'd receive the original block/inline and the resulting `HtmlElement`, allowing you to add whatever attributes you'd like.

[kzykhys/ciconia](https://github.com/kzykhys/Ciconia) uses event emitters, which seems interesting.  I've never worked with them before, but perhaps it's worth investigating whether this library should use that approach instead.

colinodell:
Maybe we could implement (or at least enable) something like the decorator functionality used by [Stack](http://stackphp.com/)?  That might also enable #35 too.

I haven't thought too deeply about this, but the idea seems intriguing enough to possibly try.

hason:
Your suggestion about decorator could be solved also via #106. The renderer is a special case of a node visitor (see https://github.com/arnaud-lb/MtHaml/blob/master/lib/MtHaml/NodeVisitor/RendererAbstract.php).

hason:
In CommonMark Attributes Extension I have created custom html renderer https://github.com/webuni/commonmark-attributes-extension/blob/master/src/HtmlRenderer.php. I wanted to created it as a decorator, but unsuccessfully.

colinodell:
@hason does the commonmark-attributes-extension package solve your need, meaning we can close this?

colinodell:
I'm going to close this for now, but we can certainly reopen it later if needed.

colinodell:
I've decided to re-open this, as there does still seem to be a need for this functionality (see #145).  I'm also warming up to this idea the more I think about it.

While I do like this approach and how it works, I [still wonder whether an alternate event-oriented approach would be better](https://github.com/thephpleague/commonmark/pull/101#issuecomment-106026529).  That would require significant changes though, so maybe we should use the proposed solution for now and revisit this later?

colinodell:
Closing in favor of #165, which is identical (but with merge conflicts resolved).

PR #103, opened 2015-06-02
Added constructor dependency injection for CommonMarkConverter

hason:
For easy customization.

### DIFF
--- a/src/CommonMarkConverter.php
+++ b/src/CommonMarkConverter.php
@@ -37,13 +37,22 @@ class CommonMarkConverter
      * Create a new commonmark converter instance.
      *
      * @param array $config
+     * @param DocParser $docParser
+     * @param HtmlRendererInterface $htmlRenderer
      */
-    public function __construct(array $config = array())
+    public function __construct(array $config = array(), DocParser $docParser = null, HtmlRendererInterface $htmlRenderer = null)
     {
-        $environment = Environment::createCommonMarkEnvironment();
-        $environment->mergeConfig($config);
-        $this->docParser = new DocParser($environment);
-        $this->htmlRenderer = new HtmlRenderer($environment);
+        $this->docParser = $docParser;
+        if (null === $docParser) {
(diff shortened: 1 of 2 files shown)


colinodell:
I've always considered `CommonMarkConverter` to be a simple wrapper for people (~95% of users) who don't need to customize anything beyond the configuration.  For the other 5%, creating three objects and tying them together is pretty trivial.  However, I do see the value in having an injectable object which ties together the parser and renderer.

I think the biggest challenge is that both the parser and renderer need to have the same `Environment` injected into them.  There are a few ways we could do this:

1. Setter injection.  Make both the parser and converter implement `EnvironmentAwareInterface` and have `CommonMarkConverter` call `setEnvironment()` on them.
2. Split the `Environment` in two - one for parsing, another for rendering - thus eliminating the need for a common class.  However, I think there is some crossover somewhere, making this impossible.
3. Create the parser/renderer in a closure, like this:

~~~php
$obj->setDocParser(function (EnvironmentInterface $environment) {
    return new MyCustomDocParser($environment);
});
~~~

But using closures doesn't seem compatible with configuration-based containers though (like Symfony's), though I could be wrong.

So while I appreciate your effort and merge request, I'm not entirely convinced this feature or approach is feasible.  I definitely welcome your thoughts and feedback though!

(Also, it seems there's a small bug with your implementation - the `$config` goes unused if a `DocParser` is provided.)

hason:
What about create a new class ``ExtendableCommonMarkConvertor``:

```php
class ExtendableCommonMarkConvertor extends CommonMarkConvertor
{
    public function __construct(DocParser $docParser, HtmlRendererInterface $htmlRenderer);
}
```

Most of the other projects use ``CommonMarkConvertor`` as API. For example, I must create a new class for those if I want to use a custom renderer.

colinodell:
Closing this as I prefer #105 (or [my alternate suggestion](https://github.com/thephpleague/commonmark/pull/105#issuecomment-109369250)) better.

PR #105, opened 2015-06-03
Added extensible converter

hason:
Alternative to #103

### DIFF
--- a/src/CommonMarkConverter.php
+++ b/src/CommonMarkConverter.php
@@ -17,22 +17,8 @@ namespace League\CommonMark;
 /**
  * Converts CommonMark-compatible Markdown to HTML.
  */
-class CommonMarkConverter
+class CommonMarkConverter extends Converter
 {
-    /**
-     * The document parser instance.
-     *
-     * @var \League\CommonMark\DocParser
-     */
-    protected $docParser;
-
-    /**
-     * The html renderer instance.
-     *
-     * @var \League\CommonMark\HtmlRendererInterface
-     */
-    protected $htmlRenderer;
-
     /**
      * Create a new commonmark converter instance.
      *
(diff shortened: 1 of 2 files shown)


colinodell:
I do like this better than #103.  Before I accept it, let me run a possible alternative by you:

**What if we move `CommonMarkConverter`'s current constructor to a static method, and then use the constructor you've provided here instead?**  So it would basically look like this:

~~~php
class CommonMarkConverter
{
    public function __construct(DocParser $docParser, HtmlRendererInterface $htmlRenderer)
    {
        $this->docParser = $docParser;
        $this->htmlRenderer = $htmlRenderer;
    }

    public function convertToHtml($commonMark)
    {
        $documentAST = $this->docParser->parse($commonMark);

        return $this->htmlRenderer->renderBlock($documentAST);
    }

    public static function create(array $config = array())
    {
        $environment->mergeConfig($config);

        $docParser = new DocParser($environment);
        $htmlRenderer = new HtmlRenderer($environment);

        return new static($docParser, $htmlRenderer);
    }
}
~~~

The downside is that it'll break BC.  Technically that's okay with `0.x` releases, but I've always intended this class to remain consistent while allowing the internals to change drastically.

The upside is that `CommonMarkConverter` will be injectable like it should be.

I'm not sure how I feel about the static method though.

Any thoughts on this?  (If you like this approach, should the `create()` method be called something better?)

/cc @philsturgeon and @GrahamCampbell - I'd greatly appreciate any guidance you might have on this.

colinodell:
I think I've decided on the direction I'd like to go (for now anyway):

1. Create a new class called `Converter` or `BaseConverter`.
2. Make its constructor accept a `DocParser` and `HtmlRendererInterface`
3. Move the `convertToHtml` method into it
4. Have `CommonMarkConverter` extend from this class
5. Have `CommonMarkConverter`'s constructor create the `Environment`, `DocParser`, and `HtmlRenderer`, passing the latter two into the parent's constructor.

This ensures backwards-compatibility and adds the functionality you're looking for in a logical manner.  It's similar to your proposed change, but with the inheritance hierarchy reversed.

@hason if you're up to making that change I will gladly accept it :)  Otherwise I'll work on it sometime this week.

hason:
@colinodell Accepted and updated.

colinodell:
Awesome, thanks @hason!

PR #106, opened 2015-06-03
Added support for element visitors

hason:
Todo:
- [x] Finish implementation
- [x] Add tests

### DIFF
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -34,15 +34,22 @@ class Converter
     protected $htmlRenderer;
 
     /**
+     * @var \League\CommonMark\ElementTraverser|null
+     */
+    protected $elementTraverser;
+
+    /**
      * Create a new commonmark converter instance.
      *
      * @param DocParser $docParser
      * @param HtmlRendererInterface $htmlRenderer
+     * @param ElementTraverser|null $elementTraverser
      */
-    public function __construct(DocParser $docParser, HtmlRendererInterface $htmlRenderer)
+    public function __construct(DocParser $docParser, HtmlRendererInterface $htmlRenderer, ElementTraverser $elementTraverser = null)
     {
         $this->docParser = $docParser;
         $this->htmlRenderer = $htmlRenderer;
+        $this->elementTraverser = $elementTraverser;
     }
 
     /**
@@ -57,6 +64,10 @@ class Converter
     public function convertToHtml($commonMark)
     {
         $documentAST = $this->docParser->parse($commonMark);
+
+        if (null !== $this->elementTraverser) {
+            $documentAST = $this->elementTraverser->traverseBlock($documentAST);
+        }
 
         return $this->htmlRenderer->renderBlock($documentAST);
     }
--- a/src/ElementTraverser.php
+++ b/src/ElementTraverser.php
@@ -1,0 +1,105 @@
+<?php
+
+/*
+ * This file is part of the league/commonmark package.
+ *
+ * (c) Colin O'Dell <colinodell@gmail.com>
+ *
+ * Original code based on the CommonMark JS reference parser (http://bitly.com/commonmark-js)
+ *  - (c) John MacFarlane
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace League\CommonMark;
+
+use League\CommonMark\Block\Element\AbstractBlock;
+use League\CommonMark\Block\Element\AbstractInlineContainer as AbstractBlockInlineContainer;
+use League\CommonMark\Block\Element\Document;
+use League\CommonMark\Inline\Element\AbstractInline;
+use League\CommonMark\Inline\Element\AbstractInlineContainer;
+
+/**
+ * ElementTraverser is a element traverser.
+ *
+ * It visits all elements and their children and calls the given visitor for each.
+ */
+class ElementTraverser
+{
+    private $environment;
+    private $context;
+
+    /**
+     * Constructor.
+     *
+     * @param Environment $environment
+     * @param ElementVisitorInterface[] $visitors
+     */
+    public function __construct(Environment $environment)
+    {
+        $this->environment = $environment;
+        $this->context = new Context(new Document(), $environment);
+    }
+
+    /**
+     * Traverses a node and calls the registered visitors.
+     *
+     * @param AbstractBlock $block
+     *
+     * @return AbstractBlock
+     */
+    public function traverseBlock(AbstractBlock $block)
+    {
+        foreach ($this->environment->getElementVisitors() as $visitor) {
+            $block = $this->traverseBlockForVisitor($visitor, $block);
+        }
+
+        return $block;
+    }
+
+    private function traverseBlockForVisitor(ElementVisitorInterface $visitor, AbstractBlock $block)
+    {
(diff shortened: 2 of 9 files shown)


hason:
@colinodell What do you think?

colinodell:
The concept sounds interesting, I'm just not sure I fully understand the purpose.  For example:

- What need does this feature address?
- How does this make the core library better?
- How does this make extensions and customizations better or easier to integrate?
- Is there a particular limitation you're experiencing with the library that led you to create this?
- Are there any similar/different approaches that might achieve the same result?

I'd like to have a clearer picture before giving feedback or making a decision on this (if that makes senses).

hason:
I want to create an extension for parsing block and inline attributes - https://pythonhosted.org/Markdown/extensions/attr_list.html The problem is with atttributes for block:
```
### A hash style header ### {: #hash }
```
The reference to the parent block is is assigned too late https://github.com/thephpleague/commonmark/blob/master/src/DocParser.php#L140 and inline processor is not the best place for addressing this task.

I want to change AST for different renderers (html page and revealjs presentation, in html page delete ``<hr>`` tags).

The node visitor can be used for optimization of AST or for completed complicated structures, as is mentioned above or in #123.

The visitor is widely accepted design pattern - twig, mthaml, rst, etc.

colinodell:
Thanks for providing that explanation!  I haven't encountered the need to do this myself, so it's helpful to see how others like yourself are using the library.

I agree, I think your visitor implementation is especially useful for that type of work.  On the other hand, it requires non-trivial changes to core functionality to add a feature which the project itself wouldn't use.

One of my guiding principles with this project is to focus on CommonMark support first and foremost. Everything that goes into this package must support that purpose. The ability to extend and customize it is a secondary goal made possible by the underlying design.

For example, the `Extension` functionality was added because we wanted to decouple the instantiation of parsers/renders from the `Environment`.  The implementation we decided on also made it possible for people to define their own custom extensions too.  Allowing multiple extensions didn't meet our primary goal, but the implementation was so trivial that it made sense to add that support anyway.

My biggest concern with your proposed change is that every `Converter`, `Environment`, and `Extension` must all know about the `ElementTraverser` now.  This is a major, non-trivial change for something the library doesn't even use itself.

That being said, I'd definitely be open to including the `ElementTraverser` (and its `ElementVisitorInterface`) as a helpful utility, perhaps until the `League\CommonMark\Util` namespace.  Anyone wishing to use this would be required to instantiate and connect their own `DocParser` and `HtmlRenderer`, putting this utility in the middle (literally three lines of code).  I think that approach would strike the right balance of allowing customization without "unnecessary" changes to the core.

What are your thoughts on this?

colinodell:
Oops, I meant to say "perhaps _inside_ the `League\CommonMark\Util` namespace" instead of "perhaps _until_ the `League\CommonMark\Util` namespace".  I've updated my comment accordingly (and will try to proof-read before submitting next time :wink:)

colinodell:
Closing due to inactivity, plus the fact that #162 may solve the underlying issue.

fesor:
I trying to implement Api Blueprint parser using this library as markdown AST provider. Currenly implementation of `InlineParserEngine` doesn't allow to use 

### Use case

Api Blueprint specification defines something like document sections. Sections declaration looks like this:

```
# <keyword> <identifier>
```

I have list of defined keywords and regular expression for identifiers. Both keyword and identifier are optional depending on section type.

I wish to implement inline parsers which will split header into this nodes, which simplifies section matching.

Without element visitors i have now only this possibilities:

 - Modify InlineParserEngine

  `InlineParserEngine` relies on `InlineElement` interface, which should provide array of possible characters. This blocks me from implementing inline parsers for identifier and for keywords parsers forces to use something like this:

  ```
return array_reduce($keywords, function ($characters, $keyword) {
    $char = substr($keyword, 0, 1);

    return array_merge($characters, [$char, strtoupper($char)]);
}, []);
```

  Since `InlineProcessorInterface` doesn't provide access to inline container, i need to replace `InlineParserEngine` implementation and modify core, as it instantiated in `DocParser` constructor.

 - Extend `Environment` class and modify inline parsers list in runtime using EnvironmentAwareInterface, which even more complicated.

colinodell:
Alright, you guys have convinced me that an element visitor is pretty much necessary.  Thanks for the all feedback and persistence :)

I think my biggest concern now would be this specific implementation.  It requires each visitor to return exactly one block/inline every time it visits something, thus allowing you to swap the block/inline with a different instance.  Does this really need to be done?  What if you don't want to swap instances, or you want to remove it, or replace it with two instances?  I'm not sure this specific implementation handles those scenarios well.

Perhaps it would be better to implement the doubly-linked list approach used by the JS reference parser?  It has a NodeWalker which can be used to traverse the AST, and then you can call methods like `insertAfter`, `insertBefore`, `appendChild`, `prependChild`, `remove`, etc directly on each node.

Would something like that solve everyone's needs?  If so, I'd strongly prefer to take that route.  It'll require a hefty rewrite of some core parts but I think the end result would be much nicer to work with.

colinodell:
(Re-opening for discussion)

fesor:
I like your suggestion, especially about insertAfter and insertBefore - this will simplify a lot of stuff (right now i have some workarounds to do such manipulations).

May I help you with that? I will have some free time next week.

colinodell:
> May I help you with that? I will have some free time next week.

Yeah, that would be awesome!

I'd recommend taking a look at both the [reference JS parser](https://github.com/jgm/commonmark.js/blob/master/lib/node.js) and [Atlassian's Java fork](https://github.com/atlassian/commonmark-java/tree/master/commonmark/src/main/java/org/commonmark/node).  I'd like to use that same type of `Node` base class for blocks and inlines.

This new node approach was introduced in commonmark.js version 0.16.  Here's the diff between 0.15 and 0.16, in case that helps: https://github.com/jgm/commonmark.js/compare/0.15...0.16  I think the specific node-related changes start around commit [e564e0a](https://github.com/jgm/commonmark.js/commit/e564e0af5c2f7b9732e89707655068e507579a89), so [this diff](https://github.com/jgm/commonmark.js/compare/e564e0a%5E...0.16) might be a little better.  (0.16 is an older version, so there's definitely changes made since then, but maybe these diffs will help you with the refactoring process.)

I started working on this a while back but got sidetracked.  I was able to find some of that code, in case it helps: https://dl.dropboxusercontent.com/u/449877/tmp/node-wip.tar.gz  You're welcome to use it, rewrite it, or ignore it completely - whatever works best for you :)

Thanks so much for taking this on, I really appreciate it and look forward to seeing what you come up with!

fesor:
> I'd like to use that same type of Node base class for blocks and inlines.

well... there is an issue with `Block/Element/AbstractInlineContainer` then... 

Inline elements could be stored as child of block element since such elements couldn't have both nested blocks and inlines. But this will cause BC break. Is it normal?

Right now i have small PoC implementation which doesn't provide any BC breaks (only small one - there is no `children` property, which affects `ListBlock`), but it's duplicates some methods.

colinodell:
I'm completely okay with breaking BC to implement this properly. In fact,
it's probably necessary, especially with those base classes.

colinodell:
Let me know if the double-linked list approach (#169) solves this issue/concern for you - if not, perhaps we can explore alternate solutions based on that.

PR #107, opened 2015-06-04
Add integrations list

jadb:
Refs https://github.com/thephpleague/oauth2-server/commit/c174b6fc65f034dd6547da6ab7bb429a20988eb0

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ Documentation can be found at [commonmark.thephpleague.com][docs].
 ## Integrations
 
 - [CakePHP 3](https://github.com/gourmet/commonmark)
+- [Laravel 5](https://github.com/GrahamCampbell/Laravel-Markdown)
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -148,6 +148,10 @@ This package is **not** part of CommonMark, but rather a compatible derivative.
 
 Documentation can be found at [commonmark.thephpleague.com][docs].
 
+## Integrations
+
+- [CakePHP 3](https://github.com/gourmet/commonmark)
+
 ## Testing
 
 ``` bash
@@ -159,7 +163,7 @@ This will also test league/commonmark against the latest supported spec.
 ## Performance Benchmarks
 
 You can compare the performance of **league/commonmark** to other popular parsers by running the included benchmark tool:
- 
+
 ``` bash
 $ ./tests/benchmark/benchmark.php
 ```


colinodell:
I have added this project to the wiki.  Thanks!

jadb:
@colinodell I am afraid not everyone goes to the wiki and other league packages are showing integrations in the README, wouldn't it be better to follow same template (look) for all the league projects? 

Thanks!

colinodell:
There's a link in the README to the relevant wiki page: https://github.com/thephpleague/commonmark/blob/master/README.md#community-extensions  Perhaps this section should be revised to make it more prominent/clear?

I'd also be open to listing them on the project website (see the [`gh-pages` branch](https://github.com/thephpleague/commonmark/tree/gh-pages)).

I'd prefer these approaches over maintaining the list within the README itself.

jadb:
I've been giving this a little more thought and I still think that if all the League packages followed the same structure for the README would be much better for end users like myself. 

I mean those packages are all using the same doc templates (gh-pages) and in general, a very standardized way of doing things, why would integrations for this particular package be listed in a different place than all the rest? The other packages might end up moving to wiki, I don't know, but standardizing it so people know where to look for all the league packages is what am trying to get to.

Maybe others can share their opinions too? /cc @philsturgeon @frankdejonge

colinodell:
The [skeleton](https://github.com/thephpleague/skeleton) template we use doesn't have an integrations section, and only some projects include this section, so it's definitely not standardized.  You're right though, providing a similar experience will be best for the users.

I think my biggest concern is that README is versioned.  It's highly possible that any "supported" integration/extension listed there may not actually support new versions we release.  OTOH, this probably isn't a big deal, and perhaps README inclusion could encourage those package authors to stay up-to-date or risk removal?

You make some good points though so I'll reconsider this and get back to you soon.  I'm open to any additional feedback you or others may have.

GrahamCampbell:
Missing https://github.com/GrahamCampbell/Laravel-Markdown for Laravel. :)

jadb:
:+1: for also including the Laravel one, want me to add to this PR @GrahamCampbell ?

GrahamCampbell:
:+1:

philsturgeon:
I like it being in the README.

README.md
```diff
@@ -148,6 +148,11 @@ This package is **not** part of CommonMark, but rather a compatible derivative.
 
 Documentation can be found at [commonmark.thephpleague.com][docs].
 
+## Integrations
+
+- [CakePHP 3](https://github.com/gourmet/commonmark)
+- [Laravel 5](https://github.com/GrahamCampbell/Laravel-Markdown)
```
GrahamCampbell: It's also laravel 5 too.

GrahamCampbell: It's also laravel 4 too.

GrahamCampbell: https://github.com/GrahamCampbell/Laravel-Markdown/tree/v2.1.0

colinodell:
Alright README it is! Thanks for being patient and persistent with me ;-)

jadb:
@colinodell @philsturgeon @GrahamCampbell thanks!

Issue #108, opened 2015-06-04
Bug with list items indented >= 4 spaces


colinodell:
See https://github.com/jgm/commonmark.js/issues/42

colinodell:
Fixed in 050b127

PR #112, opened 2015-06-10
Spec 0.20

colinodell:


colinodell:
Fixing #108 should get the tests passing.

PR #124, opened 2015-06-19
Added reference to parent of inline element

hason:


src/Inline/Element/AbstractInline.php
```diff
@@ -22,4 +24,36 @@
...
+     * @var AbstractInline|null
+     */
+    protected $parent;
+
+    /**
+     * @return AbstractBlock|AbstractInline|null
+     */
+    public function getParent()
+    {
+        return $this->parent ? : null;
```
colinodell: You could probably remove the `? : null` part, since `null` is the only value which would trigger that part of the expression anyway.

src/Inline/Element/AbstractInlineContainer.php
```diff
@@ -19,14 +19,14 @@
     /**
      * @var AbstractInline[]
      */
-    protected $children = [];
+    private $children = [];
```
colinodell: What's the rationale for making children `private` and having parents be `protected`?  Should they both be `private` or `protected`?  (I haven't thought through this - just throwing the question out there)

src/Inline/Element/AbstractInlineContainer.php
```diff
@@ -45,6 +45,9 @@ public function getChildren()
     public function setChildren($contents)
     {
         $this->children = $contents;
+        foreach ($contents as $content) {
+            $content->getParent($this);
```
colinodell: Should this be `setParent` instead?

src/Inline/Element/Image.php
```diff
@@ -27,11 +27,7 @@ public function __construct($url, $label = '', $title = '')
     {
         parent::__construct($url);
 
-        if (is_string($label)) {
-            $this->children = new ArrayCollection([new Text($label)]);
-        } else {
-            $this->children = $label;
-        }
+        $this->setChildren(is_string($label) ? new ArrayCollection([new Text($label)]) : $label);
```
colinodell: I'd prefer keeping this as an `if`/`else` for readability.

colinodell:
Thanks for submitting this enhancement!

I left some feedback on specific lines.  Would you mind reviewing those and also adding some unit tests for the new methods?

hason:
Done.

hason:
ping @colinodell

colinodell:
Merged.  Thanks for the reminder!

PR #132, opened 2015-07-06
Fix CS errors

bcrowe:


colinodell:
Thanks for fixing those!

GrahamCampbell:
@colinodell How about you enable https://styleci.io/ on this repo?

colinodell:
Oh nice, didn't know that existed!  I will definitely check that out.  Thanks for the recommendation!

Issue #98, opened 2015-05-18
Smartypants


philsturgeon:
This is probably another package, but in the same vein as the potential markua package, it would be good to get [Smartypants](http://daringfireball.net/projects/smartypants/) involved.

This would help CommonMark projects be usable for printed books and whatnot, turning standard quotes into curly quotes and other similar bits.

philsturgeon:
Although if https://github.com/thephpleague/commonmark/issues/97 happens I don't know if LaTeX/Pandoc will handle that stuff magically. Either way, would be a nice one.

cebe:
for LaTeX these should be converted not to the unicode characters but LaTeX specific syntax: http://en.wikipedia.org/wiki/Quotation_mark#Typewriters_and_early_computers

philsturgeon:
Cool. Priority goes down the list, meaning I won’t bother to do it. :D

-- 
Phil

cebe:
you need this more than me right now, so we'll see ;)

philsturgeon:
I actually don’t need it at all if LaTeX does that stuff. Just a feature suggestion now.

cebe:
not sure how this works in english, but for german quotes you need special sytnax in LaTeX. Will check it out.

philsturgeon:
Oh I see, so I _will_ need it. I just need to make sure it outputs to either HTML entities or LaTeX special syntax.

-- 
Phil

cebe:
exactly

cebe:
german wikipedia has more details about the characters to be used: http://de.wikipedia.org/wiki/Anf%C3%BChrungszeichen#Kodierung

colinodell:
Smartypants is a great idea, but I do think it should be its own separate package.  You might be able to implement it as an Inline Processor (which iterates over Inlines, performing modifications to them).  If not then we could look at expanding the "Extension API" to make these types of addons easier to implement.

cebe:
One important part is that it should not be applied to code blocks and also inline code. Will check out the API and let you know if there are limitations.

PR #134, opened 2015-07-07
Add smart punctuation

bfrohs:
Per jgm/commonmark.js@7bd852b40170f9a32864c6644ac7b5248f880f12

This PR implements the smart punctuation work done in commonmark.js and fixes a few bugs with this package.

- Smart quote, ellipsis, and en/em-dash replacement
- Punctuation tests added from jgm/commonmark.js
- Fix a unicode issue with parsing (see Environment.php)
- Fix bug where only a single Processor could be added (see closeBracketParser.php, EmphasisProcessor.php, and InlineParserEngine.php)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
     "require-dev": {
         "erusev/parsedown": "~1.0",
         "jgm/CommonMark": "0.20",
+        "jgm/SmartPunct": "0.20",
         "michelf/php-markdown": "~1.4",
         "phpunit/phpunit": "~4.3",
         "phpunit/phpunit-mock-objects": "2.3.0",
@@ -36,6 +37,17 @@
                 "version": "0.20",
                 "dist": {
                     "url": "http://spec.commonmark.org/0.20/spec.txt",
+                    "type": "file"
+                }
+            }
+        },
+        {
+            "type": "package",
+            "package": {
+                "name": "jgm/SmartPunct",
+                "version": "0.20",
+                "dist": {
+                    "url": "https://raw.githubusercontent.com/jgm/commonmark.js/0.20.0/test/smart_punct.txt",
                     "type": "file"
                 }
             }
--- a/src/Inline/Parser/SmartPunctParser.php
+++ b/src/Inline/Parser/SmartPunctParser.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the league/commonmark package.
+ *
+ * (c) Colin O'Dell <colinodell@gmail.com>
+ *
+ * Original code based on the CommonMark JS reference parser (http://bitly.com/commonmark-js)
+ *  - (c) John MacFarlane
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace League\CommonMark\Inline\Parser;
+
+use League\CommonMark\ContextInterface;
+use League\CommonMark\InlineParserContext;
+use League\CommonMark\Inline\Element\Text;
+
+class SmartPunctParser extends AbstractInlineParser
+{
+    /**
+     * @return string[]
+     */
+    public function getCharacters()
+    {
+        return ['-', '.'];
+    }
+
+    /**
+     * @param ContextInterface $context
+     * @param InlineParserContext $inlineContext
+     *
+     * @return bool
+     */
+    public function parse(ContextInterface $context, InlineParserContext $inlineContext)
+    {
+        $cursor = $inlineContext->getCursor();
+        $ch = $cursor->getCharacter();
+
+        // Ellipses
+        if ($ch === '.' && $matched = $cursor->match('/^\\.( ?\\.)\\1/'))
+        {
+            $inlineContext->getInlines()->add(new Text("…"));
+            return true;
+        }
+
+        // Em/En-dashes
+        elseif ($ch === '-' && $matched = $cursor->match('/^(?<!-)(-{2,})/'))
+        {
+            $count = strlen($matched);
+            $en_dash = '–';
+            $en_count = 0;
+            $em_dash = '—';
+            $em_count = 0;
+            if ($count % 3 === 0) {
+                $em_count = $count / 3;
+            } elseif ($count % 2 === 0) {
+                $en_count = $count / 2;
+            } elseif (($count - 2) % 3 === 0) {
+                $em_count = floor(($count - 2) / 3);
+                $en_count = 1;
+            } else {
+                $em_count = floor(($count - 4) / 3);
+                $en_count = 2;
+            }
+            $inlineContext->getInlines()->add(new Text(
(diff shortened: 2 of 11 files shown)


colinodell:
Wow, thanks so much for taking this on!  On first glance this looks pretty good.  Would it be possible to split those fixes into separate commits?  It would make it a little easier to visualize and manage those changes.

I'll take a closer look later tonight or tomorrow and get back to you with any other thoughts or comments.  Thanks again!

bfrohs:
> Would it be possible to split those fixes into separate commits? It would make it a little easier to visualize and manage those changes.

Done. Let me know if you need anything else!

cebe:
nice, so this basially implements #98 :)

colinodell:
Oh good call @cebe, I forgot about that one.

I think wrapping this up as an `Extension` would make it even easier to use.  The question becomes whether that should be included in this package or perhaps as a separate (yet still) official package?

bfrohs:
I've amended the most recent commit to remove an unnecessary line (adding a `Quote` element which didn't turn out to be necessary in the end and I forgot to clean it up).

> I think wrapping this up as an `Extension` would make it even easier to use. The question becomes whether that should be included in this package or perhaps as a separate (yet still) official package?

I think it would make sense in the main repo, since it's part of jgm/commonmark.js. Were you thinking something along the lines of c2c9ec9? It is a barebones extension that, if used by itself, only parses punctuation and paragraphs. Otherwise, it can be added to another environment to inherit from there (as I updated the tests to do).

Just let me know if you want to keep it or not and I can change/remove it as necessary.

cebe:
I have no strong opinion on whether this should be a separate package or not, it is not a heavy one and the JS implementation also ships it in one package so it may just stay in this repo.

colinodell:
@bfrohs yes that is exactly what I had in mind! 

@cebe correct, this would not be enabled by default.  And I agree with your point, perhaps it's best to just include this specific feature.

I want to do one final read-through of the code tonight and then I will accept the merge. :smile: 

On a related side-note - I'm also wondering if perhaps we should look at reorganizing the classes a bit; basically separating the core parser, CommonMark elements/parsers, and this feature into 3 separate sub-namespaces.  Right now things are all lumped in together.  It's fine for now though - just something to think about in the future, especially if other similar features might be coming.

bfrohs:
@colinodell Fixed the latest commit per https://github.com/thephpleague/commonmark/commit/c2c9ec9b9a502681b0a9b430e999ea5301c5b795#commitcomment-12068861

src/Inline/Processor/QuoteProcessor.php
```diff
@@ -0,0 +1,46 @@
...
+                $closerInline->getContent() === '“' ? '”' : '’'
+            );
+
+            return $closer->getNext();
+        };
+
+        // Process the emphasis characters
+        $delimiterStack->iterateByCharacters(['“', "’"], $callback, $stackBottom);
+        // Remove gaps
+        $inlines->removeGaps();
```
colinodell: Lines 43 and 44 can (should?) probably be removed.  I'm thinking this was probably copypasted from the other processor which basically nulls-out some of the inlines to mark them for eventual removal by `removeGaps()`.  We're not doing this here, so the `removeGaps()` call shouldn't be necessary.

src/Inline/Parser/SmartPunctParser.php
```diff
@@ -0,0 +1,77 @@
...
+            if ($count % 3 === 0) {
+                $em_count = $count / 3;
+            } elseif ($count % 2 === 0) {
+                $en_count = $count / 2;
+            } elseif (($count - 2) % 3 === 0) {
+                $em_count = floor(($count - 2) / 3);
+                $en_count = 1;
+            } else {
+                $em_count = floor(($count - 4) / 3);
+                $en_count = 2;
```
colinodell: Lines 62-66 [aren't covered by any tests](https://scrutinizer-ci.com/g/thephpleague/commonmark/inspections/7783b039-5521-4e49-8f53-36a7899ad4b4/code-structure/class/League%5CCommonMark%5CInline%5CParser%5CSmartPunctParser).  Are you sure they are needed?  If so, we may want to include a test for them.

colinodell:
Alright I've taken a close look and am really happy with what you've done :)  I've just got two final comments.  Let me know your thoughts.

src/Inline/Processor/QuoteProcessor.php
bfrohs: Yup, copy/paste issue. Fixed.

src/Inline/Parser/SmartPunctParser.php
bfrohs: They are needed; I did manual testing on another project to make sure. I don't have time to add the tests right now, but I'll try to add tonight or tomorrow.

colinodell: Awesome! And if you could make sure those also match [the reference JS parser](http://spec.commonmark.org/dingus/)'s output that would be perfect.

bfrohs: As much as it pains me to change this behavior, I've done so to match commonmark.js. However, we may need to revert the commit in the future depending on jgm/commonmark.js#56. Anyway, should be all set for now!

composer.json
```diff
@@ -23,6 +23,7 @@
     "require-dev": {
         "erusev/parsedown": "~1.0",
         "jgm/CommonMark": "0.20",
+        "jgm/SmartPunct": "0.20",
```
hason: Why? The ``smart_punct.txt`` file is included in ``jgm/CommonMark``.

colinodell:
Merged.  Thanks so much for this contribution!

src/Inline/Parser/SmartPunctParser.php
colinodell: Ah okay.  I must admit, I didn't realize the reference parser's dash handling wasn't ideal.  I'd definitely prefer to use the same rules as the reference parser though, so I agree with bringing it up there.  If that discussion doesn't lead to the change you want to see, I'd be open to implementing your strategy anyway, but let's exhaust this option first.

composer.json
bfrohs: > Why? The smart_punct.txt file is included in jgm/CommonMark.

Is it? `jgm/CommonMark` points to <http://spec.commonmark.org/0.20/spec.txt> in [`composer.json`](https://github.com/thephpleague/commonmark/blob/master/composer.json#L36), which doesn't seem to include any smart punctuation tests. Additionally, there is no `smart_punct.txt` file in `jgm/CommonMark` anywhere I can find it.

There is, however a [`smart_punct.txt` file in `jgm/CommonMark.js`](https://github.com/jgm/commonmark.js/blob/master/test/smart_punct.txt), which is what `jgm/SmartPunct` points to in this file.

hason: I'm sorry, I didn't notice it.

PR #136, opened 2015-07-08
Fix wording for invalid inline exception

bfrohs:
This has thrown me off a few times. Looks like a copy/paste issue. (Same exact error message can be found later in the file for invalid blocks.)


colinodell:
Yep, good call!

PR #139, opened 2015-07-10
Switch behavior back to "smart" hyphen replacement per jgm/commonmark.js#58

bfrohs:
jgm/commonmark.js#58 was merged with the behavior originally proposed here. jgm/cmark has also been updated to match.

This code passes current tests and it also passes the new tests that aren't part of a commonmark.js release yet (tested manually).


colinodell:
Awesome!  Thanks so much for getting those implemented upstream too!

PR #142, opened 2015-07-14
Fixed crazy indentation

GrahamCampbell:


colinodell:
Thanks for fixing that!

PR #147, opened 2015-07-17
Added reference to parent block for inline elements before processing

hason:
I want to solve a hack with ``debug_backtrace`` in https://github.com/webuni/commonmark-attributes-extension/blob/master/src/AttributesInlineProcessor.php#L39

### DIFF
--- a/src/DocParser.php
+++ b/src/DocParser.php
@@ -137,7 +137,7 @@ class DocParser
     {
         if ($block instanceof AbstractInlineContainer) {
             $cursor = new Cursor(trim($block->getStringContent()));
-            $block->setInlines($this->inlineParserEngine->parse($context, $cursor));
+            $block->setInlines($this->inlineParserEngine->parse($context, $cursor, $block));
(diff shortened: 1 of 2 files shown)


colinodell:
Aren't we already setting the parent of these inlines inside the [setInlines() method](https://github.com/thephpleague/commonmark/blame/master/src/Block/Element/AbstractInlineContainer.php#L57-59)?

colinodell:
I'd prefer only setting the parent relationship once.

Also, I don't really like how we have to modify two methods to pass this information into `InlineParserEngine::processInlines`.  Perhaps we should pass it as part of the `InlineParserContext` instead.  Or maybe there's a different/better way to achieve what you're looking for?

hason:
My problem would be solved by implementing the visitor patter as I suggested in #106.

colinodell:
Closing this in favor of #162.  It's basically the same approach, just done slightly differently.  Thanks for providing the idea!

Issue #143, opened 2015-07-15
Spec 0.21 support


colinodell:
- https://github.com/jgm/CommonMark/releases/tag/0.21
- https://github.com/jgm/commonmark.js/releases/tag/0.21.0

PR #148, opened 2015-07-18
Spec 0.21

colinodell:
Implements the new spec changes (#143)


colinodell:
Unfortunately things got a bit slower with these changes: https://blackfire.io/profiles/compare/554d8c88-bc17-405b-b92b-38e56ed25656/graph

PR #151, opened 2015-07-19
Updated styleci config

GrahamCampbell:


GrahamCampbell:
Our recommended fixer set is very similar to symfony, only actually matches what you wanted better.

GrahamCampbell:
There are a few places where you've put `return null;` at the end of a function. You should actually delete those statements entirely since they do absolutely nothing. Both the symfony and recommended reset fixed those to `return;`.

colinodell:
Thanks, I'll give this a try!

Issue #133, opened 2015-07-06
Automate code style checks


colinodell:
Perhaps use https://styleci.io/ as suggested by @GrahamCampbell [here](https://github.com/thephpleague/commonmark/pull/132#issuecomment-118845310)?

GrahamCampbell:
We support configuration too, so if you don't like our default checks, you can tell us to just use PSR-2 and nothing extra.

colinodell:
@GrahamCampbell How do I disable the end-every-comment-with-a-period rule?  (Is there a list of configuration settings readily available somewhere?)

GrahamCampbell:
> end-every-comment-with-a-period rule

Disable `phpdoc_short_description`. :)

GrahamCampbell:
The current version of StyleCI uses 1.9.1 of https://github.com/FriendsOfPHP/PHP-CS-Fixer.

Soon we're moving to 2.0-dev though. We'll notify all our users 5 days before we upgrade.

GrahamCampbell:
I'd recommend you don't disable the rule though since short descriptions are meant to be actual sentences, so are required to start with a capital letter and end with a full stop.

colinodell:
I really like the ruleset you suggested.  Thanks so much for helping with this!

GrahamCampbell:
No problem. :)

PR #152, opened 2015-07-19
Style fixes (#133)

colinodell:


src/Delimiter/DelimiterStack.php
```diff
@@ -178,6 +178,6 @@ protected function findFirstMatchingOpener(Delimiter $closer, $openersBottom, De
             $opener = $opener->getPrevious();
         }
 
-        return null;
+        return;
```
GrahamCampbell: This statement should be removed entirely.

PR #154, opened 2015-07-19
Improved phpdoc

GrahamCampbell:


colinodell:
I think there's an extra space on the `@return string ` line.

GrahamCampbell:
I'll pull this down and do it on my PC in a min.

PR #161, opened 2015-07-19
Reduce nesting and remove useless returns

GrahamCampbell:


colinodell:
Good calls!

GrahamCampbell:
You might want to disable scrutinizer for pull requests. It seems to take ages and do nothing.

PR #162, opened 2015-07-19
Set inline parents before processing them

colinodell:
It's possible that inlines may be added/modified, so we'll still need to set the parents
afterwards (this functionality already exists).

This is an alternate implementation to #147.


GrahamCampbell:
How does this compare performance wise?

GrahamCampbell:
Ie, how much slower does this make things?

colinodell:
It looks like it's a 6% performance penalty.  Not too bad, but not ideal either.

Perhaps I can just make the container block available to the processors should they need it, instead of setting the parents everywhere on the off-chance that a processor might need it.

colinodell:
Hmm, I'll have to think on this.  Perhaps I'll end up implementing #106 after all?

PR #163, opened 2015-07-20
Fix replacement of multiple apostrophes in the same paragraph

bfrohs:
Per jgm/commonmark.js#61, which isn't yet merged, but don't see why it wouldn't be.

This could probably be merged right away, if you wanted, since it doesn't break any existing tests. The test it helps to pass is in jgm/commonmark.js#61.


colinodell:
Merged.  Thanks for identifying and fixing that!

PR #164, opened 2015-07-24
Change double quote algorithm to match spec

bfrohs:
See jgm/commonmark.js@ddd2f5e3adc0e29aa354837a5e4b3014dfff3311


colinodell:
Thanks!

PR #165, opened 2015-07-25
Added support for tag attributes in renderers

colinodell:
This is identical to #101, but with the merge conflicts resolved.

Full credit goes to @hason for implementing this - thanks a ton!

### DIFF
--- a/src/Inline/Element/AbstractInline.php
+++ b/src/Inline/Element/AbstractInline.php
@@ -59,4 +59,15 @@ abstract class AbstractInline
 
         return $this;
     }
+
+    /**
+     * @param string     $key
+     * @param mixed|null $default
+     *
+     * @return mixed|null
+     */
+    public function getData($key, $default = null)
+    {
+        return array_key_exists($key, $this->data) ? $this->data[$key] : $default;
+    }
 }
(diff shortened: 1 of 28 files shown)


src/Inline/Element/AbstractInline.php
```diff
@@ -59,4 +59,15 @@ public function setParent($parent)
 
         return $this;
     }
+
+    /**
+     * @param string     $key
+     * @param mixed|null $default
```
GrahamCampbell: mixed|null doesn't really make sense since null is a subset of mixed

src/Inline/Element/AbstractInline.php
```diff
@@ -59,4 +59,15 @@ public function setParent($parent)
...
+     *
+     * @return mixed|null
```
GrahamCampbell: same here

src/Inline/Element/AbstractInline.php
colinodell: Good point.  I think I'm using that elsewhere, so I'll replace all such instances.

gavin-how:
Hi, 
I was hoping someone could help in helping me understand how to use this? 
What is the syntax for putting attributes in Markdown? Or am I misunderstanding what this is.

Cheers
Gavin

PR #167, opened 2015-07-25
Remove unnecessary trim from ListItemRenderer

colinodell:
This ensures our renderer returns an HtmlElement instead of a string, which is easier to work with


GrahamCampbell:
:+1:

GrahamCampbell:
How far are we off a 0.10 release btw?

colinodell:
0.10 should be ready in a few minutes hopefully :)  I just want to take a final look through all the issues/PRs and make sure I'm not missing anything critical.

GrahamCampbell:
Awesome. :)

PR #169, opened 2015-07-28
WIP: Double-linked list based AST

fesor:
- [x] Add base `Node` object which uses double-linked list
- [x] Cleanup deprecated methods calls
- [x] Add NodeWalker
- [x] Simplify inline parsers
- [x] Cleanup code

---

This PR includes major refactoring of current commonmark implementation

## Changes

- Double-linked list based AST
- Simplified `Delimiter`s and Inline parsers: instead of storing position, we now can store InlineNode.
- Ability to modify AST in runtime by using NodeWalkers
- `AbstractInlineContainer` replaced with `InlineContainer` interface

### NodeWalker
NodeWalker can be used to change AST in runtime. Each node has `walker` method, which will create node walker for given node. Here is example of how to capitalize all emphasis in document:

```php
$doc = 'Hello *world*';

$ast = $parser->parse($doc);
$walker = $ast->walker();
$inEmphasis = false;

while (($event = $walker->next())) {
    $node = $event->getNode();
    if ($node instanceof Emphasis) {
        if ($event->isEntering()) {
            $inEmphasis = true;
        } else {
            $inEmphasis = false;
            // add Emphasis node's children as siblings
            while($node->firstChild()) {
                $node->insertBefore($node->firstChild());
            }
            // remove the empty Emphasis node
            $node->detach();
        }
    } else if ($inEmphasis && $node instanceof Text) {
        $node->setContent(strtoupper($node->getContent()));
    }
}

$html = $renderer->renderBlock($ast); // <p>Hello WORLD</p>
```


colinodell:
Hey @fesor,

Sorry for the delay in getting back to you, this week has been crazy.

I think you're off to a great start!  I like how you're starting with a mostly-BC approach, as that'll probably make the next refactoring steps easier.  However, I'm totally okay with breaking BC completely for this specific feature.  I'd rather get this right for a 1.0 release instead of being stuck halfway there (between the current implementation and a node-based one).

I think I like the Node Walker approach best.  It's simple yet powerful, and it easily supports custom node types.  It's also from the JS library which we should be trying to adhere to as much as possible.

So yeah, I'm really happy with the direction this is going.  I'd just ask that you break BC to make the implementation cleaner and closer to the JS library's node implementation.  The closer we can match it the easier it'll be to pull down upstream changes in the future.

Thanks so much for taking this on!

Colin

fesor:
Node Walker is ready. This is example of how to use it (taken from commonmark.js):

Capitalize all emphasis 
```
$doc = 'Hello *world*';

$ast = $parser->parse($doc);
$walker = $ast->walker();
$inEmphasis = false;

while (($event = $walker->next())) {
    $node = $event->getNode();
    if ($node instanceof Emphasis) {
        if ($event->isEntering()) {
            $inEmphasis = true;
        } else {
            $inEmphasis = false;

            // add Emphasis node's children as siblings
            while($node->getFirstChild()) {
                $node->insertBefore($node->getFirstChild());
            }

            // remove the empty Emphasis node
            $node->detach();
        }
    } else if ($inEmphasis && $node instanceof Text) {
        $node->setContent(strtoupper($node->getContent()));
    }
}

$renderer->renderBlock($ast); // <p>Hello WORLD</p>
```

fesor:
Well i think that this PR is too big... double linked lists possible simplifies:

 - inline parsing (in progress right now)
 - inline processors
 - block parsing (allows to move parsing details from nodes to parsers)
 - rendering (one interface for both inline and block nodes)

I think it would be better to split this PR... To do so we need some definition of done to be declarated...

fesor:
@colinodell ping

colinodell:
Sorry, I've been swamped with work.  I will definitely take a look at this
either tonight or tomorrow.  Thanks for being patient (and for the gentle
reminder haha)

fesor:
Hi, it's ok, i understand (i have the same problem). I just need your comments to know what to do next... I really not sure about some of my changes (especially changes in `InlineContextInterface`).

colinodell:
I took a quick look just now - I think you're still on the right track!  Most of these changes make complete sense to me.

I'll need some more time to review these changes in more depth, especially the `InlineParserContext`.  I can't recall exactly why I put the `$inlines` collection there in the first place, so I'll need to figure that out before can I say whether your change works well.

My goal is to fully review and tinker with these changes in the next day or two.  I'll send a more-detailed reply once that's done.  If everything looks good, perhaps we can merge this PR as-is and then create new PRs for the additional changes?  I'd be okay with that approach if you are.

Thanks again for your patience and persistence, this is looking great! :)

src/Node/Node.php
```diff
@@ -0,0 +1,232 @@
...
+     * @return Node[]
+     */
+    public function getChildren()
+    {
+        $children = [];
+        for ($current = $this->firstChild;null !== $current;$current = $current->next) {
+            $children[] = $current;
+        }
+
+        return $children;
```
colinodell: Perhaps we could eventually use a generator here?  I'm fine with creating a temporary array for now, it shouldn't be too cumbersome.

src/Block/Element/AbstractBlock.php
```diff
@@ -84,75 +74,35 @@ public function __construct()
      */
     public function getParent()
     {
-        return $this->parent;
+        return parent::getParent();
```
colinodell: Could we simply remove the `getParent` method from this class?

src/Inline/Element/AbstractInline.php
```diff
@@ -26,38 +26,11 @@
...
-                'League\CommonMark\Block\Element\AbstractBlock',
-                'League\CommonMark\Block\Element\AbstractInline',
-                is_object($parent) ? get_class($parent) : gettype($parent)
-            ));
-        }
-
-        $this->parent = $parent;
-
-        return $this;
+        return false;
```
colinodell: Some inlines (like emphasis and links) **are** containers (https://github.com/jgm/commonmark.js/blob/0.21.0/lib/node.js#L3-L19)

I believe this `isContainer()` check is currently only used in the [`NodeWalker::next()` method](https://github.com/thephpleague/commonmark/pull/169/files#diff-0cfd46ad72d87b09256e734b343b397aR39).

Here's an idea - what if we create a `ContainerNode` which extends from `Node`, and move the following methods (and their related variables) into it:

* `appendChild()`
* `prependChild()`
* `getChildren()`
* `getFirstChild()`
* `getLastChild()`
* `detachChildren`
* `replaceChildren`

Then, in `NodeWalker`, replace this line:

~~~php
if ($entering && $current->isContainer()) {
~~~

With:

~~~php
if ($entering && $current instanceof ContainerNode) {
~~~

Therefore avoiding the need to define a separate `isContainer` method - you simply look at the object type.

What are your thoughts on that?

colinodell: Oops, I missed that `AbstractInlineContainer` was returning `true` - sorry about that.

Even so, perhaps it's still a good idea?

src/Inline/Element/AbstractInlineContainer.php
```diff
@@ -44,9 +39,9 @@ public function getChildren()
      */
     public function setChildren($contents)
     {
-        $this->children = $contents;
-        foreach ($contents as $content) {
-            $content->setParent($this);
+        $this->detachChildren();
+        foreach ($contents as $children) {
+            $this->appendChild($children);
```
colinodell: This seems to do the same thing as `replaceChildren`, so I'd be okay with removing this method and using the base `replaceChildren` method instead.

Actually, I think we could probably remove this class altogether and use `ContainerNode` instead.  Thoughts?

src/Inline/Parser/CloseBracketParser.php
```diff
@@ -89,26 +82,29 @@ public function parse(ContextInterface $context, InlineParserContext $inlineCont
...
         }
 
+        $node = $this->createInline($link['url'], $link['title'], $isImage);
+        $openerNode = $opener->getInlineNode();
+        $openerNode->insertBefore($node);
+        while (($inline = $openerNode->getNext()) !== null) {
+            $node->appendChild($inline);
+        }
+        $openerNode->detach();
+
```
colinodell: Would you mind describing these changes for me?  It might be quicker for you to explain them than for me to reason them out myself :)

src/Inline/Processor/EmphasisProcessor.php
```diff
@@ -35,30 +35,28 @@ public function processInlines(ArrayCollection $inlines, DelimiterStack $delimit
...
+                $emph = new Strong();
             }
 
-            // Insert into list of inlines
-            $inlines->set($opener->getPos() + 1, $emph);
-            for ($i = $opener->getPos() + 2; $i < $closer->getPos(); $i++) {
-                $inlines->set($i, null);
+            $openerInline->insertAfter($emph);
+            while (($node = $emph->getNext()) !== $closerInline) {
+                $emph->appendChild($node);
```
colinodell: I really like how the `Node`-based approach works here :)

src/DocParser.php
```diff
@@ -132,15 +133,19 @@ private function incorporateLine(ContextInterface $context)
...
+                continue;
+            }
 
-        foreach ($block->getChildren() as $child) {
-            $this->processInlines($context, $child);
+            $node = $event->getNode();
+            if ($node instanceof InlineContainer) {
+                $context->setContainer($node);
+                $cursor = new Cursor(trim($node->getStringContent()));
+                $this->inlineParserEngine->parse($context, $cursor);
```
colinodell: Prior to this PR, the **only** reason we pass the `$context` in is because [`CloseBrackParser` needs access to the document's reference map](https://github.com/thephpleague/commonmark/blob/master/src/Inline/Parser/CloseBracketParser.php#L84).  This seems like a bad architectural decision on my part - we should only pass along what we need, and we certainly don't need the whole `$context` here.  (The `Context` class is primarily designed for tracking the state while handling *blocks*, whereas `InlineParserContext` is designed to track state when dealing with *inlines*.)

Your PR removes `$inlineContext->getInlines()` and uses `$context->getContainer()` instead.  This makes us even more reliant on the `$context`, which I'd like to avoid if possible.

Therefore, if we wanted to avoid using the block-oriented `$context` when parsing inlines, we'd need to make the following changes:

## 1. Add new properties to the `InlineParserContext`

We'll want to add `Node $container` and a `ReferenceMap $refMap` to the constructor plus the corresponding properties and getter methods.

## 2. Rewrite the `InlineParserEngine::parse` method

Change the method from this:

~~~php
 /**
     * @param ContextInterface $context
     * @param Cursor           $cursor
     *
     * @return ArrayCollection
     */
    public function parse(ContextInterface $context, Cursor $cursor)
    {
        $inlineParserContext = new InlineParserContext($cursor);
        /// ....
    }
~~~

To this:

~~~
 /**
     * @param Node         $container
     * @param ReferenceMap $refMap
     *
     * @return ArrayCollection
     */
    public function parse(Node $container, ReferenceMap $refMap)
    {
        $inlineParserContext = new InlineParserContext($node, $refMap);
        // ...
    }
~~~

And have the `InlineParserContext` create the `Cursor` it needs from the node, instead of [relying on an external caller to create that](https://github.com/thephpleague/commonmark/pull/169/files#diff-64d43afca3fac22bce1c4d9a3b84310aR146).

## 3. Remove the `ContextInterface $context` parameter from `AbstractInlineParser::parse()` and its subclasses

The method would change from:

~~~php
public function parse(ContextInterface $context, InlineParserContext $inlineContext)
~~~

To:

~~~php
public function parse(InlineParserContext $inlineContext)
~~~

The container would be obtained via `$inlineContext->getContainer()` and the reference map would be obtained via `$inlineContext->getReferenceMap()`.

## 4. Update the above `DocParser::processInlines` method

The code inside  the `while` loop would instead look like this:

~~~php
$node = $event->getNode();
if ($node instanceof InlineContainer) {
    $this->inlineParserEngine->parse($node);
}
~~~

----

I think this approach greatly simplifies and decouples things.  What are your thoughts?

If you like this approach, would you like to implement these changes or would you prefer that I do it?  Either is fine with me - just let me know :)

colinodell:
@fesor I found some time tonight to review the changes in depth.  I've commented on your code with some thoughts and questions - if you could review those I'd appreciate it.

Regarding the `InlineContextInterface` changes: I think we should probably have that class store the container and the reference map, which would completely eliminate our dependency on the `Context` class while handling inlines.  Here's a longer, in-depth comment on that: https://github.com/thephpleague/commonmark/pull/169/files#r36934703

I'm really excited about this PR and look forward to getting it merged in soon!

src/DocParser.php
fesor: Yes, I like this approach, will implement this today.

src/Node/Node.php
fesor: I thought about that, but this will require to drop php 5.4 support. Is it OK?

src/Block/Element/AbstractBlock.php
fesor: Yes, ...I think this method overridden only because of phpdoc

src/Inline/Element/AbstractInlineContainer.php
fesor: Well, this method used only because of `Image` and `Link` requires label nodes to be passed to constructor. So yes, I can just drop this method.

src/Inline/Parser/CloseBracketParser.php
fesor: Well, firstly we create link or image node, and then we fill it with label nodes. After that we remove opener node.

src/Inline/Processor/EmphasisProcessor.php
fesor: :)

src/Inline/Element/AbstractInline.php
fesor: I really don't remember why I drop `NodeContainer`... I will think about that.

src/Node/Node.php
colinodell: Good point.  Let's keep this method as-is for now.

src/Block/Element/AbstractBlock.php
colinodell: If we implement the `ContainerNode` idea then perhaps we can make the `Node::getParent()` docblock be `@return ContainerNode`.

colinodell:
> Right now ReferenceMap instantiated in Documents's constructor. Possible solution is to move it out from document and pass to it's constuctor. Any thoughts?

What is the benefit of making this change?  I know that passing dependencies via the constructor is a good practice, but does this particular case have any other advantages for us?

fesor:
@colinodell , this is related to your comment

> We'll want to add Node $container and a ReferenceMap $refMap to the constructor plus the corresponding properties and getter methods.

When InlineParserEngine created, there is no document and RefMap.

src/Inline/Element/AbstractInline.php
fesor: Well.. I had some concerns about `AbstractInlineContainer` (it already inherits from `AbstractInline`).

colinodell:
> When InlineParserEngine created, there is no document and RefMap.

[Your implementation](https://github.com/fesor/commonmark/commit/72e7d3b7dd208d97a57f2261999e8afd432845d8#diff-86ad2fee84a59082e7bd864c735b64a0) is exactly what I had wanted :)

src/Inline/Element/AbstractInline.php
colinodell: What if we remove `AbstractInline` and `AbstractBlock`?  Block and inline elements could then extend directly from `Node` or `ContainerNode`, depending on which is appropriate.

We could probably remove `AbstractInlineContainer` too.

fesor:
My bad :)

src/Inline/Element/AbstractInline.php
fesor: Well... the we can't check whether type of node is block or inline. Also block nodes have some parsing-specific methods (which could be moved to parsers... i planned to do it in another PR).

I will think about it.

fesor: Well, i tried to implement `ContainerNode` and failed... there is no way implement it without multiple inheritance... May be you have some thoughts about this problem?

Currently the problem is that every inline object already inherited from `Node`. Possible solution is to drop `AbstractInline` and `AbstractBlock` classes and make everything just a node. What do you think?

src/Inline/Parser/CloseBracketParser.php
fesor: This could be simplified to:

```php
$node = $this->createInline($link['url'], $link['title'], $isImage);
$opener->getInlineNode()->replaceWith($node);
while (($inline = $node->getNext()) !== null) {
      $node->appendChild($inline);
}
```

colinodell: I think I prefer your simplified code - it's a little bit clearer to me.

src/Inline/Element/AbstractInline.php
colinodell: > there is no way implement it without multiple inheritance... May be you have some thoughts about this problem?

Yeah, I encountered this issue myself a long time ago when I first tried making these changes.

Maybe we could replace `AbstractInline` and `AbstractBlock` with interfaces?  For example:

~~~php
class Node {
    // ...
}

class ContainerNode extends Node {
    public function appendChild() { }
    public function prependChild() { }
    public function getChildren() { }
    public function getFirstChild() { }
    public function getLastChild() { }
    public function detachChildren() { }
    public function replaceChildren() { }
}

interface BlockElementInterface {
    // All of the abstract functions currently found in AbstractBlock:
    public function canContain(Node $node);
    public function acceptsLines();
    public function isCode();
    public function matchesNextLine();
    public function handleRemainingContents();
}

interface InlineElementInterface {
    // TODO: Is this interface even necessary?
}

// This replaces AbstractStringContainer
interface StringContainerInterface extends InlineElementContainer {
    public function getContent();
    public function setContent();
}

// This replaces AbstractWebResource
interface WebResourceInterface extends InlineElementContainer {
    public function getUrl();
}

~~~

(I have removed the method parameters above to keep the example simple)

You'd therefore end up combining them like so:

~~~php
class Paragraph extends ContainerNode implements BlockElementInterface { }

class HorizontalRule extends Node implements BlockElementInterface { }

class Emphasis extends ContainerNode implements InlineElementInterface { }

class Newline extends Node implements InlineElementInterface { }

class Text extends Node implements StringContainerInterface { }
~~~

If you need to know whether something is a block or inline, check for the relevant interface.

Do you think something like that would work?

fesor: I thought about it, i think that this is the only acceptable approach. Will check this today.

src/Inline/Parser/CloseBracketParser.php
fesor: Or even...

```php
$inline = $this->createInline($link['url'], $link['title'], $isImage);
$opener->getInlineNode()->replaceWith($inline);
while (($label = $inline->getNext()) !== null) {
      $inline->appendChild($label);
}
```

I'll update it as soon as possible

fesor:
Well... I added PoC implementation of `NodeContainer`... This code requires review.

fesor:
There is a lot of issues with types now... see scrutinizer report. This is something that should be fixed, but i don't know how. Most of the issues was introducing by adding `BlockElement`.

colinodell:
Thanks for working on the PoC implementation of `NodeContainer`!

I see what you mean about the Scrutinizer errors, and I understand why they're occurring.  The more I think about this, the more I feel that `NodeContainer` isn't a great idea after all.  Other AST implementations (including the JS reference parser) don't differentiate between different node types - a node is simply a node.

The `NodeContainer` idea sounded great in theory, and I do like the changes you made.  However, I'm now worried that perhaps we're unintentionally creating more problems than we're solving.  What do you think about removing/undoing that PoC, thus reverting back to having a single type of `Node`?

fesor:
Yeh, I would prefer to revert this changes. I don't like the idea of  `NodeContainer`, since we have nodes of a tree and this is logical to not separate this things.

But i also don't like the `isContainer` method. In nodejs implementation `isContainer` is separate function, in which nodes are listed as is. Possible way is to allow to specify via `Environment` which node is container and which isn't. But I don't know.

Also a have some other thoughts about some things (like removing parsing specific stuff from block elements). Well before we can get some solution (after reverting PoC implementation), what can I do is... add tests and cleanup stuff.

colinodell:
> Yeh, I would prefer to revert this changes. I don't like the idea of NodeContainer, since we have nodes of a tree and this is logical to not separate this things.

Makes sense to me!

> But i also don't like the isContainer method. In nodejs implementation isContainer is separate function, in which nodes are listed as is. Possible way is to allow to specify via Environment which node is container and which isn't (this may be used to add more flexibility). But I don't know.
>
> Also a have some other thoughts about some things (like removing parsing specific stuff from block elements, or cleaning up attributes handling). Anyway before we can get some solution (after reverting PoC implementation), what can I do is... add tests and cleanup stuff.

Those are some really good points, and I'd definitely be interested in hearing your thoughts on other things.  But yeah, let's go ahead and finalize this PR first and we'll address those other things next :)

colinodell:
@fesor Do you think this is ready to be merged?  I don't mind if we add the extra unit tests as a future PR.

fesor:
Well... I think it is. There is some issues with this PR but they can be solved in future PRs.

fesor:
Maybe i should summarize changes in this PR, or something else?

colinodell:
I've merge in the changes :) Thanks a ton for working on this!

I'll go ahead and update the CHANGELOG.

Tomorrow I'm leaving for a week-long vacation, so I won't be around to review or accept any PRs, but feel free to work on whatever changes or improvements you'd like.  Or take a break from this project yourself :)

colinodell:
@fesor I've documented the changes here: https://github.com/thephpleague/commonmark/commit/e275f14bbf3adee4ffdfea914223198373d771a2  Feel free to open a PR if I missed anything.

Issue #172, opened 2015-08-24
0.22.0 spec


colinodell:
https://github.com/jgm/commonmark.js/releases/tag/0.22.0

We'll want to test against this new spec and port over any changes or bug fixes.  This may need to wait until #169 is merged though (to avoid conflicts).

PR #174, opened 2015-08-28
[WIP] Spec 0.22

colinodell:
See https://github.com/jgm/commonmark.js/compare/0.21.0...0.22.0 and #172

We'll need to check which of these changes should be ported into here:

- [x] Fix logic error in calculation of offset
- [x] Disallow list item starting with multiple blank lines
- [x] Fix bug in list-start parser

### DIFF
--- a/src/Cursor.php
+++ b/src/Cursor.php
@@ -193,7 +193,6 @@ class Cursor
         while ($advanceByColumns ? ($cols < $characters) : ($i < $characters)) {
             if ($this->peek($i) === "\t") {
                 $cols += (4 - (($this->column + $cols) % 4));
-
             } else {
                 $cols++;
             }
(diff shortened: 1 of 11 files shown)


colinodell:
When I try implementing [this fix](https://github.com/jgm/commonmark.js/commit/f3b7f162a71a25558fd3d848ef40e1d3aac0c990) it ends up breaking test #260.  I took a quick look but couldn't immediately find the cause.  So far I've confirmed that the list block is marked as tight, which is incorrect.  It might be related to determining whether lines are blank, or perhaps adding new paragraph blocks - I'm not sure.

I'm currently on vacation, so if anyone could take a look at this I'd really appreciate it!

src/Cursor.php
```diff
@@ -192,7 +192,8 @@ public function advanceBy($characters, $advanceByColumns = false)
         $cols = 0;
         while ($advanceByColumns ? ($cols < $characters) : ($i < $characters)) {
             if ($this->peek($i) === "\t") {
-                $cols += (4 - ($this->column % 4));
+                $cols += (4 - (($this->column + $cols) % 4));
+
```
GrahamCampbell: cs

Issue #122, opened 2015-06-19
Review & update gh-pages


colinodell:
It's been a while since I touched the docs.  We should probably revisit them and make sure they're still up-to-date.

colinodell:
Also add dynamic changelog (see https://github.com/thephpleague/glide/commit/f7307cdddeb54a20355b43731186450a536343ad#diff-a4cb471ed14a06b17e38fc51d8e7744c and https://github.com/thephpleague/skeleton/issues/33)

colinodell:
Docs will definitely need an update once #169 is merged in.

colinodell:
I've made some updates to the docs, and I think they're up-to-date now.  If somebody else wants to take a second look I'd appreciate it :)

PR #178, opened 2015-09-13
Initial updated docs commit

postalservice14:
Part of issue: #122

I couldn't figure out how Glide's gh-pages works without a *.md file in the root-level directory.  I couldn't get this to work without copying `/0.10/index.md -> /index.md`.  Then it worked great.


colinodell:
Thank you for working on this!  I'll take a closer look at it sometime this week (hopefully tomorrow or Tuesday).

colinodell:
@postalservice14,

Thanks for taking the time to start working on this!

The CommonMark spec isn't stable yet, so this library could conceivably undergo several major changes before a 1.0 release.  Perhaps 0.11-0.12 would be similar, but then 0.13 and 0.14 could differ drastically.  I'd therefore prefer to maintain a single set of docs for now, especially since we're still in the 0.x phase.

We'll definitely want to reconsider this once we reach 1.0, 2.0, etc. though!  Separate docs with a version selector would be super important then.

I'll definitely keep this PR bookmarked for the future :)  Thanks again for getting this started!

postalservice14:
Sure, no problem.  I'll keep the branch on my fork available if you want to reference it in the future.

PR #179, opened 2015-09-14
Make `AbstractBlock::handleRemainingContents()` no longer abstract

colinodell:
A default implementation is provided.  This was based on code which was duplicated across several
sub-classes.  This duplicate code is therefore unneeded and was removed.

### DIFF
--- a/src/Block/Element/BlockQuote.php
+++ b/src/Block/Element/BlockQuote.php
@@ -14,7 +14,6 @@
 
 namespace League\CommonMark\Block\Element;
 
-use League\CommonMark\ContextInterface;
 use League\CommonMark\Cursor;
 
 class BlockQuote extends AbstractBlock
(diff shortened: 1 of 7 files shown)


src/Block/Element/BlockQuote.php
```diff
@@ -67,21 +66,6 @@ public function matchesNextLine(Cursor $cursor)
     }
 
     /**
-     * @param ContextInterface $context
-     * @param Cursor           $cursor
-     */
-    public function handleRemainingContents(ContextInterface $context, Cursor $cursor)
-    {
-        if ($cursor->isBlank()) {
```
GrahamCampbell: isn't this stuff lost though?

colinodell: Nope, because of [this change from #174](https://github.com/thephpleague/commonmark/commit/752a6a4c398515af15380c322dcc5b16bf289a8a#diff-64d43afca3fac22bce1c4d9a3b84310aR126):

~~~diff
 if ($context->getContainer()->isOpen()) {
-    if ($context->getContainer()->isOpen()) {
+    if ($context->getContainer()->acceptsLines()) {
         $context->getContainer()->handleRemainingContents($context, $cursor);
~~~

We were checking the wrong method (`isOpen()` instead of `acceptsLines()`).  This resulted in `handleRemainingContents()` being called in cases where it shouldn't have been.  This would have caused issues, except that `if ($cursor->isBlank()) { return; }` intervened and aborted the function in these cases.

In other words, these particular instances resulted in no-ops before and will continue to do so now (since they won't even be called), so we're okay :)

Good eye though!

GrahamCampbell: Ah, great. ;)

PR #181, opened 2015-09-19
Add `./bin/commonmark` for converting Markdown at the command line

colinodell:
TODO:

 - [x] Fix code style issues
 - [x] Remove `.php` from filename
 - [x] Add some tests
 - [x] Add documentation to gh-pages

### DIFF
--- a/bin/commonmark
+++ b/bin/commonmark
@@ -1,0 +1,112 @@
+#!/usr/bin/env php
+<?php
+
+requireAutoloader();
+
+ini_set('display_errors', 'stderr');
+
+foreach ($argv as $i => $arg) {
+    if ($i === 0) {
+        continue;
+    }
+
+    if (substr($arg, 0, 1) === '-') {
+        switch ($arg) {
+            case '-h':
+            case '--help':
+                help();
+                break;
+            default:
+                fail('Unknown option: ' . $arg);
+        }
+    } else {
+        $src = $argv[1];
+    }
+}
+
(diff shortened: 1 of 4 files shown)


bin/commonmark
```diff
@@ -0,0 +1,112 @@
...
+
+    $markdown = file_get_contents($src);
+} else {
+    $stdin = fopen('php://stdin', 'r');
+    stream_set_blocking($stdin, false);
+    $markdown = stream_get_contents($stdin);
+    fclose($stdin);
+
+    if (empty($markdown)) {
+        help();
```
cebe: help output may better go to STDERR instead of STDOUT to have empty output on stdout when input is empty.

bin/commonmark
```diff
@@ -0,0 +1,112 @@
...
+{
+    $autoloadPaths = [
+        // Local package usage
+        __DIR__ . '/../vendor/autoload.php',
+        // Package was included as a library
+        __DIR__ . '/../../../vendor/autoload.php',
+    ];
+    foreach ($autoloadPaths as $path) {
+        if (file_exists($path)) {
+            require_once $path;
```
cebe: should `break` after first one has been included.

colinodell: Good catch, thanks!

bin/commonmark
colinodell: Fixed, thanks for the advice!

PR #184, opened 2015-09-23
Fix #183

### LINKED ISSUE #183: LogicException: You cannot add lines to a block which cannot accept them
Noticed this when upgrading from 0.10.0 to 0.11.1.

# Exception trace

src/Block/Element/AbstractBlock.php in addLine at line 233

    throw new \LogicException('You cannot add lines to a block which cannot accept them');

src/Block/Element/IndentedCode.php in handleRemainingContents at line 94

    $context->getTip()->addLine($cursor->getRemainder());

src/DocParser.php in incorporateLine at line 133

    $context->getContainer()->handleRemainingContents($context, $cursor);

src/DocParser.php in parse at line 86

    $this->incorporateLine($context);

src/Converter.php in convertToHtml at line 59

    $documentAST = $this->docParser->parse($commonMark);

# Minimal test case

```
10. Bullet

        code


Test
```

colinodell:
This should fix #183 (at least temporarily, if not permanently)

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ The following table shows which versions of league/commonmark are compatible wit
     </thead>
     <tbody>
         <tr>
-            <td><strong>0.11.1</strong><br>0.11.0</td>
+            <td><strong>0.11.x</strong></td>
             <td><strong><a href="http://spec.commonmark.org/0.22/">0.22</a></strong></td>
         </tr>
         <tr>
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -126,7 +126,7 @@ The following table shows which versions of league/commonmark are compatible wit
     </thead>
     <tbody>
         <tr>
-            <td><strong>0.11.1</strong><br>0.11.0</td>
+            <td><strong>0.11.x</strong></td>
```
GrahamCampbell: should be a 2 shouldn't it?

colinodell: So far all three `0.11` versions are compatible (and the next one should be too) so I'm taking the lazy approach here :P

GrahamCampbell: Right. :)

PR #185, opened 2015-09-23
Reset container after closing containing lists (#183)

colinodell:


colinodell:
Because this is more-or-less functionally equivalent to the 0.11.2 fix, I'm probably going to hold off an issuing a new release until we have something else to include as well.  (We'll still merge this into `master` though.)

GrahamCampbell:
I see you made a release anyway. :)

PR #197, opened 2015-10-19
Update travis-ci to thephpleague's skeleton style

postalservice14:
This removes the need to download ocular.phar (using wget) within travis.  Also adds the ability to check the composer version constraints with (--prefer-lowest).

But all in all, follows the skeleton projects design. (https://github.com/thephpleague/skeleton)


colinodell:
I like it, thank you for the PR!

PR #205, opened 2015-11-16
Fixed SmilieParser example

ikkentim:
This or should be an and in order for the parser to work correctly.

### DIFF
--- a/customization/inline-parsing.md
+++ b/customization/inline-parsing.md
@@ -100,7 +100,7 @@ class SmilieParser extends AbstractInlineParser
         // The next character must be a paren; if not, then bail
         // We use peek() to quickly check without affecting the cursor
         $nextChar = $cursor->peek();
-        if ($nextChar !== '(' || $nextChar !== ')') {
+        if ($nextChar !== '(' && $nextChar !== ')') {
             return false;
         }
 


GrahamCampbell:
This is wrong.

GrahamCampbell:
The next character can't be two characters at the same time...

colinodell:
@GrahamCampbell is correct.  The goal of this `if` statement is to ensure the next character is either a `(` or `)`.  If the character is something else, we abort immediately.

Thanks for taking a look though!

luukholleman:
No @ikkentim is right, this statement is always true. If the character is ), then it's not ( and so the statement is true. Which will return false.

ikkentim:
Sorry, but @GrahamCampbell and @colinodell are both mistaken.

I didn't expect to have to explain this obvious issue.

```
-        if ($nextChar !== '(' || $nextChar !== ')') {
+        if ($nextChar !== '(' && $nextChar !== ')') {
```

If we're looking whether the next character is NOT ( and NOT ) then my PR is correct.

If we're using the old code ( with the `||` ). The next would happen:

Lets say we're processing `:(`. The next character is `(`. The if statement would basically be like `if( true || false )` which is true, meaning the code will return `false` and not use the parser. 

Lets say we're processing `:)`. The next character is `)`. The if statement would basically be like `if( false || true )` which is true, meaning the code will return false and not use the parser. 

Basically, `parse` will always return `false`, which is wrong.

colinodell:
Yep, you're right, my sincerest apologies.  (I didn't have my coffee yet this morning...)

This demonstrates that the current code is indeed wrong: https://3v4l.org/XOcS0

Changing the `||` to `&&` does resolve the issue: https://3v4l.org/HjJla

Thank you for the explanation and the PR!

PR #215, opened 2016-01-13
Add optional `Environment` parameter to `CommonMarkConverter` constructor

colinodell:
This change allows developers to convert Markdown using a custom environment without needing to manually instantiate the doc parser and renderer.


colinodell:
I'm not a big fan of having this parameter come after the config array, but it's the easiest way to maintain my BC promise for this class.

I also don't particularly like how the unit test checks a protected property, but hey it works.

PR #221, opened 2016-01-15
Updated alt three emoji description

GrahamCampbell:
Laravel markdown has been modified to support "extensions", and alt three emoji now ships with an extension class.


colinodell:
Awesome!

PR #223, opened 2016-01-16
Simplify some logic

colinodell:


src/Inline/Parser/EmphasisParser.php
```diff
@@ -105,13 +130,13 @@ public function parse(InlineParserContext $inlineContext)
 
         $leftFlanking = $numDelims > 0 && !$afterIsWhitespace &&
             !($afterIsPunctuation &&
-            !$beforeIsWhitespace &&
-            !$beforeIsPunctuation);
+                !$beforeIsWhitespace &&
```
GrahamCampbell: this is very hard to read

PR #225, opened 2016-01-19
Fix empty regex

colinodell:


colinodell:
I accidentally branched off the wrong thing.  Let's try this again...

PR #229, opened 2016-01-22
Update inline-rendering.md

Arcesilas:
Fixed typo


colinodell:
Thanks for catching and fixing that! I really appreciate it :)

PR #235, opened 2016-02-07
Updated styleci config

GrahamCampbell:
This fixer has been renamed: https://styleci.readme.io/docs/fixers#phpdoc_summary. Though the alias will still work, using the new name is preferred.


colinodell:
Awesome, thanks!

PR #238, opened 2016-03-24
Convert URLs that support https but are http presently

anthonyryan1:
Excludes tests to avoid reducing coverage.

This is a little bit pedantic, but I forgot to exclude my vendor directory from a grep for insecure http URLs and saw these. Figured I'd sent a pull request to convert the ones that support it.


colinodell:
I'll happily accept a positive change, no matter how small or pedantic it may seem :) Thanks for the PR!

PR #239, opened 2016-03-25
Enable calling convertToHtml as a function

### LINKED ISSUE #233: Add __invoke() method to Converter
This would allow using the `Converter` or `CommonMarkConverter` objects as functions:

```
$converter = new CommonMarkConverter();
echo $converter($markdown);
```

See https://github.com/thephpleague/html-to-markdown/pull/85

healsdata:
Adding Converter::__invoke as an alias to Converter::convertToHtml
to enable calling Converter object as functions

resolves #233

### DIFF
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -60,4 +60,18 @@ class Converter
 
         return $this->htmlRenderer->renderBlock($documentAST);
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


colinodell:
Thanks for implementing this!

PR #248, opened 2016-05-31
Remove version constraint to let Composer figure out the best version

mnapoli:
That way you are sure your doc is not outdated. As a user I didn't want to copy/paste that command because I'm not sure if it's up to date, and I can get the same result with a simpler command.


colinodell:
I do keep it up-to-date, but you're right, this is much better.  Thanks!

PR #250, opened 2016-06-14
bugfix for vendor autoload.php file path

ctrl-f5:


GrahamCampbell:
I don't think this is right?

colinodell:
I think it is right.  When installed as a Composer dependency you end up with these files:

```
├── composer.json
└── vendor
    ├── autoload.php
    ├── bin
    │   └── commonmark -> ../league/commonmark/bin/commonmark
    └── league
        └── commonmark
            └── bin
                └── commonmark
```

To access `autoload.php` from `vendor/league/commonmark/bin/commonmark`, you need to back out three levels to where `autoload.php` exists.

I merged [a similar fix for league/html-to-markdown](https://github.com/thephpleague/html-to-markdown/commit/9d8116e7033c3e9c030c6f2f0096b7e62ac91433) but didn't apply that here as well.

Thanks for contributing the fix!

GrahamCampbell:
:+1:

PR #254, opened 2016-06-28
Add tests for the safe mode

mnapoli:
While having a look I noticed some tests could be useful, especially if this behavior might change later.


colinodell:
Thank you for adding these!

PR #262, opened 2016-09-14
Prepare for 0.15 release

colinodell:


colinodell:
It looks like the Windows tests of the CLI "binary" are failing :cry:

colinodell:
From what I can tell, the binary works just fine, but the dev dependency which runs the commands is not working properly.  I'm therefore going to accept this PR as-is and fix the Windows tests later.

PR #264, opened 2016-09-29
Fix inline rendering example

kelunik:


colinodell:

kelunik:
:-)

PR #272, opened 2016-12-10
Allow inline parsers matching regex delimiter to be created

### LINKED ISSUE #271: Cannot create inline parser matching forward slash character
When creating an inline parser:

```php
class ForwardSlashParser extends AbstractInlineParser
{
    public function getCharacters()
    {
        return ['/'];
    }

    public function parse(InlineParserContext $inlineContext)
    {
         return true;
    }
}
```

Then attempting to convert some input:

```php
$environment->addInlineParser(new ForwardSlashParser());

$converter = new CommonMarkConverter([], $environment)

$output = $converter->convertToHtml($input);
```

You get:

```
ErrorException in Cursor.php line 359:
preg_match(): Unknown modifier ']'
```

This can be fixed by changing line 514 Environment.php from this:

```php
$this->inlineParserCharacterRegex = '/^[^' . preg_quote(implode('', $chars)) . ']+/u';
```

to this:

```php
$this->inlineParserCharacterRegex = '/^[^' . preg_quote(implode('', $chars), '/') . ']+/u';
```

ntzm:
Fixes #271

### DIFF
--- a/tests/unit/EnvironmentTest.php
+++ b/tests/unit/EnvironmentTest.php
@@ -238,6 +238,21 @@ class EnvironmentTest extends \PHPUnit_Framework_TestCase
         $this->assertContains($parser, $environment->getInlineParsers());
     }
 
+    public function testInlineParserCanMatchRegexDelimiter()
+    {
+        $environment = new Environment();
+
+        $parser = $this->getMock('League\CommonMark\Inline\Parser\InlineParserInterface');
+        $parser->expects($this->any())
+            ->method('getCharacters')
+            ->will($this->returnValue(['/']));
+
+        $environment->addInlineParser($parser);
+        $environment->getInlineParsers();
+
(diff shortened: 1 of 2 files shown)


tests/unit/EnvironmentTest.php
```diff
@@ -238,6 +238,21 @@ public function testAddInlineParserAndGetter()
...
+
+        $parser = $this->getMock('League\CommonMark\Inline\Parser\InlineParserInterface');
+        $parser->expects($this->any())
+            ->method('getCharacters')
+            ->will($this->returnValue(['/']));
+
+        $environment->addInlineParser($parser);
+        $environment->getInlineParsers();
+
+        $this->assertEquals(1, preg_match($environment->getInlineParserCharacterRegex(), 'foo/bar'));
```
colinodell: This works for me :)

colinodell:
Thanks for the fix!

Issue #275, opened 2017-03-08
Allow external links


PauBenetPrat:
Would it be possible to have a setting 

i.e.  `allow_external_links` as like as we have `allow_unsafe_links`.

If set true all links would be converted with target=_blank attribute.

This might be better than writting `<a href="url" target=_blank >some text</a>`

Thanks

nrobinaubertin:
You can create an externalLinkRenderer like this :
```php
class ExternalLinkRenderer implements InlineRendererInterface
{
    private $host;

    public function __construct($host)
    {   
        $this->host = $host;
    }   

    public function render(AbstractInline $inline, ElementRendererInterface $htmlRenderer)
    {   
        if (!($inline instanceof Link)) {
            throw new \InvalidArgumentException('Incompatible inline type: ' . get_class($inline));
        }   

        $attrs = array();

        $attrs['href'] = $htmlRenderer->escape($inline->getUrl(), true);

        if (isset($inline->attributes['title'])) {
            $attrs['title'] = $htmlRenderer->escape($inline->data['title'], true);
        }   

        if ($this->isExternalUrl($inline->getUrl())) {
            $attrs['target'] = '_blank';
        }   

        return new HtmlElement('a', $attrs, $htmlRenderer->renderInlines($inline->children()));
    }   

    private function isExternalUrl($url)
    {   
        return parse_url($url, PHP_URL_HOST) != NULL && parse_url($url, PHP_URL_HOST) !== $this->host;
    }   
}
```
Just add it to the environment and you're good to go :  
```php
$environment = Environment::createCommonMarkEnvironment();
$environment->addInlineRenderer('League\CommonMark\Inline\Element\Link', new ExternalLinkRenderer($_SERVER["HTTP_HOST"]));
```

PauBenetPrat:
Thanks! This seems to work, however I'm using [Laravel-Markdown](https://github.com/GrahamCampbell/Laravel-Markdown) and I can not use this solution. That's why I created a pull request refs #276 ...

In that way just modifying a config setting everything works as expected

GrahamCampbell:
> Thanks! This seems to work, however I'm using Laravel-Markdown and I can not use this solution. 

Yes you can use this solution.

GrahamCampbell:
My markdown package was specifically designed so that you can customize the environment.

PauBenetPrat:
[...]

PR #276, opened 2017-03-08
Allow external links refs #275

PauBenetPrat:


PauBenetPrat:
Hi @colinodell , @marijnvdwerf, @hasonn. I don't understand why first test is not passing. What should I do.

PR #288, opened 2017-08-20
PHP 7.2 now available via Travis

colinodell:
(supposedly)


.travis.yml
```diff
@@ -8,7 +8,7 @@ php:
   - 5.6
   - 7.0
   - 7.1
-  - master
+  - 7.2
```
cebe: you should still test against master, which is the upcoming php 7.3 ;)

PR #290, opened 2017-10-30
Removed deprecated getMock and setExpectedException methods

phpanarchist:


carusogabriel:
As we need to keep compatibility with `PHPUnit 4`, `createMock` shouldn't be used. Use `getMockBuilder($class)->getMock()` instead.

tests/unit/CommonMarkConverterTest.php
```diff
@@ -37,7 +37,7 @@ public function testConfigOnlyConstructor()
     public function testEnvironmentAndConfigConstructor()
     {
         $config = ['foo' => 'bar'];
-        $mockEnvironment = $this->getMock('League\CommonMark\Environment');
+        $mockEnvironment = $this->createMock('League\CommonMark\Environment');
```
carusogabriel: As we need to keep compatibility with `PHPUnit 4`, `createMock` shouldn't be used. Use `getMockBuilder($class)->getMock()` instead.

tests/unit/EnvironmentTest.php
```diff
@@ -91,7 +91,7 @@ public function testSetConfig()
 
     public function testSetConfigAfterInit()
     {
-        $this->setExpectedException('RuntimeException');
+        $this->expectException('RuntimeException');
```
carusogabriel: As we need to keep `PHPUnit 4` compatibility, `expectException` shouldn't be used. Use `@expectedException` annotation instead.

colinodell:
I'll be dropping support for PHPUnit 4 in a few minutes :)

Thanks for the contribution!

PR #302, opened 2017-12-30
Improve Cursor performance when dealing with very long lines.

sebholl:
Came across this easy-win when parsing text that had an incredibly long JSON string embedded in it.

CommonMark assumes UTF8 encoding in strings and so makes use of the multi-byte `mb_` functions for string interrogation. When using character offsets, `mb_*` functions handling variable-width encodings (like UTF8) typically have to scan the string every time from the beginning which can have `O(n)` time complexity even if you're pulling out just 1 character. As such, all of the `getCharacter()` random access calls that CommonMark makes can cause the parser have `O(n^2)` running time. As CommonMark currently parses line-by-line, this becomes noticable on text with super loooonnngg lines.

For fixed-width encodings like ASCII, the `mb_*` functions can use `O(1)` pointer arithmetic. ASCII is basically the 1-byte codepoints of UTF8 so if we know the UTF8 string only contains ASCII codepoints, then we know that all characters are 1 byte long, and `mb_` can treat the string as if it was ASCII. To allow this, `mb_detect_encoding()` determines whether the characters are ASCII or UTF8 at the point where a `Cursor` is instantiated and this encoding is persisted and used for any subsequent calls to `mb_`.

Note: in the off-chance the string is neither ASCII or UTF8, we fall back to LATIN1 (ISO-8859-1) for which `mb_` effectively treats each byte in the string as 1 character. It's a safe, best-effort fallback.

For performance testing, I ran `benchmark.php` with a new `.md` file that I generated containing a single 64kb line of "Lorem dolor ipsum...", and the parse time went down from 6.6ms to 5.3ms.

I couldn't see how you wanted to automate performance testing but let me know if you'd like me to include this new file as part of the benchmark test going forward.

### DIFF
--- a/src/Cursor.php
+++ b/src/Cursor.php
@@ -59,12 +59,18 @@ class Cursor
     private $partiallyConsumedTab = false;
 
     /**
+     * @var string
+     */
+    private $encoding;
+
+    /**
      * @param string $line
      */
     public function __construct($line)
     {
         $this->line = $line;
-        $this->length = mb_strlen($line, 'utf-8');
+        $this->encoding = mb_detect_encoding($line, 'ASCII,UTF-8', true) ?: 'ISO-8859-1';
+        $this->length = mb_strlen($line, $this->encoding);
     }
 
     /**
@@ -175,7 +181,7 @@ class Cursor
             return;
         }
 
-        return mb_substr($this->line, $index, 1, 'utf-8');
+        return mb_substr($this->line, $index, 1, $this->encoding);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


colinodell:
Thank you for submitting this improvement!  I really appreciate the detailed description as well - that all makes sense to me.  I profiled the current benchmark using Blackfire and can confirm it does indeed shave some time off the Cursor calls.

I'm okay with keeping the current benchmark test as-is - the goal of that is mostly to see how CommonMark compares to other libraries when parsing "typical" Markdown content.  Think of it more like a quick, high-level overview.  For smaller optimizations and/or edge cases like a 64kb line of Lorem Ipsum I'll instead use Blackfire to get a detailed look at the individual method calls and confirm that performance does improve where expected.

Thanks again for the PR!  :dolphin:

sebholl:
Thank-you so much for the speedy merge, @colinodell, and your ongoing work on this library. Happy holidays!

PR #313, opened 2018-03-13
Solidify official Drupal integration project

markcarver:
Both of these projects are merging. See https://www.drupal.org/project/markdown/issues/2952435


colinodell:
That's great to hear!  Thanks for submitting this change :+1:

markcarver:
My pleasure :D Just glad that we've finally gotten our s!*# together haha

PR #317, opened 2018-03-25
benchmark things

krakjoe:
* improvements to output of benchmark script
  * add krakjoe/cmark (without hard dep)


colinodell:
This is awesome, thanks Joe!

PR #319, opened 2018-03-26
0.29 spec changes

colinodell:
See https://github.com/commonmark/commonmark.js/compare/0.28.1...master

### DIFF
--- a/src/Block/Parser/SetExtHeadingParser.php
+++ b/src/Block/Parser/SetExtHeadingParser.php
@@ -81,7 +81,7 @@ class SetExtHeadingParser implements BlockParserInterface
             }
         }
 
-        return \array_filter($strings, function($s) {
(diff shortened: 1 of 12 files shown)


src/Block/Parser/SetExtHeadingParser.php
```diff
@@ -46,8 +47,42 @@ public function parse(ContextInterface $context, Cursor $cursor): bool
         $level = $match[0][0] === '=' ? 1 : 2;
         $strings = $context->getContainer()->getStrings();
 
+        $strings = $this->resolveReferenceLinkDefinitions($strings, $cursor->getEncoding(), $context->getReferenceParser());
+        if (empty($strings)) {
```
glensc: isn't just `if (!$strings)` more readable?

PR #320, opened 2018-03-26
more benchmark improvements

krakjoe:
* selectively benchmark parsers
  * optional process isolation
  * set markdown source from cmdline
  * set iterations from cmdline
  * allow reading markdown from stdin

Maybe update readme also ...


colinodell:
I'm loving these changes, thanks for submitting them!

PR #323, opened 2018-04-01
add csv output for graphs

krakjoe:
* csv output for  graphs
  * improve isolation a little

`php benchmark.php -csv benchmark.csv`

This will do the normal test, but also write the results to the given csv file, for whatever you like ...

Supporting file descriptors means we can support piping the csv to another program, possibly gnuplot, but it's a mystery to me ...

[made with output csv](https://plot.ly/~krakjoe/3/)


colinodell:
Nice!

PR #331, opened 2018-09-01
Add rel="noopener  noreferrer" tag incase target tag is set to _blank and the rel tag is missing

azjezz:
by not passing the rel="noopener" tag, [you might allow the target page to take control of the original page](https://mathiasbynens.github.io/rel-noopener/).

so why not secure by default ? 😃


colinodell:
Thanks for the contribution!

PR #335, opened 2018-10-30
bumped phpstan up to `-l max`, should fix #334

### LINKED ISSUE #334: v2 Roadmap
My current wishlist for v2:

- [ ] AST stuff split into its own namespace or package
- [ ] Block and inline elements split into their own package (to enable sharing with league/html-to-markdown)? this is probably a "duplicate" of the AST split above
- [ ] Refactor how data is stored on AST elements; maybe use the `data` array more heavily?
- [ ] Extract some of the parsing logic out of the block/inline elements and into the parsers (similar to Atlassian's Java fork) and re-evaluate #222
- [x] Drop PHP 5.6 support
- [x] Parameter/return types everywhere
- [ ] Two-way XML conversion
- [ ] Review other CommonMark implementations for other improvement ideas
- [x] Pass phpstan checks at a higher level
- [x] ~~See whether stream/generator inputs make sense~~ (they don't)
- [ ] Other big-ticket items like #186, #232, and #312 
- [ ] We sometimes call `AbstractBlock` elements on variables which contain `Node`. These are virtually always `AbstractBlock` elements though. See if we need to move some methods between these and/or tighten up our type hints.

I'll gradually refine this list as progress and decisions are made and spin it off into a milestone.  Just wanted a place to collect my thoughts for now.

SignpostMarv:


SignpostMarv:
the `return null` styleci issues will either need to be fixed in styleci & ignored in phpstan, or ignored in style & left "fixed" in phpstan.

SignpostMarv:
the travis failures seem to be down to something in travis timing out rather than test failures?

colinodell:
I'm getting the same behavior when I run PHPUnit locally.  It's possible that one of the changes in here has resulted in an infinite loop - perhaps the `Cursor` is stuck trying to parse the same character over and over and fails to advance to the next one?

If you're able to identify and resolve that I'll go ahead and re-test this.

Once that's done I'll also run our benchmark and code coverage checks so we can properly weigh these improvements against any possible downsides.  I absolutely appreciate your contribution and understand the rationale behind it, I just need to make sure we're balancing everything :)

Thanks!

SignpostMarv:
fixed the phpunit issue, also attempted to resolve the issues identified by scrutinizer that I introduced.

colinodell:
Thanks for fixing those test failures!

I ran the benchmark (`tests/benchmark/benchmark.php`) against your branch and it does seem to be 34% slower :worried:  Version `0.18.0` runs in 21.12ms on my machine but this branch takes 28.43ms.

I also ran it through Blackfire which showed a 22% performance penalty, mostly due to a couple key functions: https://blackfire.io/profiles/compare/27460615-0e0e-42ed-8a74-220aa94a9b79/graph

Unfortunately a 22-34% performance penalty is a bit too large to merge this as-is.

I noticed you also added a method to the `ContextInterface` which ensures that the tip is always an `AbstractBlock` - was that because phpstan said it could be a different class, or was that more to protect against it being `null`?  I'd like to better understand the rationale for this change in case there's something else we also need to consider.

SignpostMarv:
I'll take a peek at the performance issues.

re: the contextinterface method, scrutinizer complained about the additional if statements introduced- shifting the condition into the context class keeps the additional if conditions for checking for `T` instead of letting `T|null` through should put the scrutinizer score back where it was.

SignpostMarv:
It looks like the performance has been impacted by the variety of additional type checks made when Node|null gets passed to a method expecting AbstractBlock|null or AbstractBlock, which is going to be tricky to resolve without sacrificing type safety vs. phpstan levels.

Might be best to roll back the phpstan level to a point where the benchmarks don't get affected then wait until php 5.6 is dropped to revisit (different syntactical solutions etc.)?

SignpostMarv:
with 5.6 going EOL in the new year, what're the thoughts on retaining support for php 5.6 in here?

colinodell:
Hey @SignpostMarv, I'm really sorry about the delay in getting back to you.

> It looks like the performance has been impacted by the variety of additional type checks made when Node|null gets passed to a method expecting AbstractBlock|null or AbstractBlock, which is going to be tricky to resolve without sacrificing type safety vs. phpstan levels.

As much as I love phpstan and like the idea of meeting the `max` level, I'm definitely more concerned about performance.  For that reason I think we should remove those extra type checks and dial back the level to something lower.

> Might be best to roll back the phpstan level to a point where the benchmarks don't get affected then wait until php 5.6 is dropped to revisit (different syntactical solutions etc.)?

I agree 100%.  I don't have a timeline for dropping PHP 5.6 yet, though.

colinodell:
Due to the performance issues, failing tests, and conflicts caused by newer changes, I'm going to close this PR.

@SignpostMarv I sincerely appreciate your work on this, but at this time I'd like to stay with our current level.  We can absolutely revisit this closer towards the 0.19 release once we're done making additional breaking changes.  I can't promise that we'll increase it, but we'll definitely consider it.  I'll create a new issue so we don't forget!

PR #343, opened 2019-03-10
github template: fix heavy copy paste

glensc:
refs:
- d54b2f0af75409b7c7cb919a85e4ef7bc4275f61

cc:
- @colinodell


colinodell:
Good catch!

PR #348, opened 2019-03-13
Use PHP 5.5 class constant

glensc:
This project minimum php requirement is 5.6, so it's time to use `::class` constant.

cc @colinodell

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "colinodell/commonmark-php": "*"
     },
     "require": {
-        "php": ">=5.6.5",
+        "php": "^5.6.5||^7.0",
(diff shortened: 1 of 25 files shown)


colinodell:
I like it!

glensc:
@colinodell complete, ready for review and merge

used this mega sed, but still need to manually review the changes in IDE

```
$ git ls-files | xargs grep "'League.CommonMark" -l|xargs sed -re "s/'(League.*)'/\1::class/g"
$ git ls-files | xargs grep "'League..CommonMark" -l|xargs sed -re "s/'(League.*)'/\\\\\1::class/g" -i
```

PR #355, opened 2019-04-01
security: add recommendation to html sanitizing

glensc:
you recommended some library to sanitize html using whitelisted tags, attributes.

but i can't find what you recommended, or where...

### DIFF
--- a/security.md
+++ b/security.md
@@ -36,6 +36,8 @@ echo $converter->convertToHtml('<script>alert("Hello XSS!");</script>');
 
 **Failing to set this option could make your site vulnerable to cross-site scripting (XSS) attacks!**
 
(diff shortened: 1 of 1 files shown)


colinodell:
Implemented via 9e03020c684a76ddb51c115bc2b642650b428743

glensc:
btw, htmlpurifier is not good. it seems old, it can't parse html5. for example i tried to make it whitelist `<details>` tag, and even with very complex code to do so, i still failed to achieve that. also could not find a way to preserve html comments.

- https://github.com/glensc/eventum/commit/92c039f4d7b4ad440ca804bd35946bca3937ddcd
- https://github.com/glensc/eventum/commits/experiment/htmlpurifier

PR #364, opened 2019-05-25
Drop support for non-UTF-8-compatible encoding (fixes #361)

### LINKED ISSUE #361: ISO-8859-1 support is broken
```
    ini_set('default_charset', 'ISO-8859-1');
    use League\CommonMark\Environment;

    $markdown = new CommonMarkConverter([], $environment);


    $text = 'foo gå `bling`-bar';

    $text = mb_convert_encoding($text, 'ISO-8859-1');

    var_dump($markdown->convertToHtml($text));
```

The above code returns 

```
<p>foo gå `.bling`-bar</p>
```

Seems UTF-8 is now required as input?

Version: 0.19.1

colinodell:
We used to claim support for ISO-8859-1 encoding, and several methods hinted at us supporting other encodings like UTF-16 too.  But in reality, we were making several assumptions and optimizations that the text was UTF-8 or ASCII.  This led to issues like #361.

As a result, and because UTF-8 is overwhelming popular, we're going to drop support for all non-UTF-8-compatible encodings.  Anyone wishing to use those can still do so, they'll just need to convert to/from UTF-8 when using this library.

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -11,8 +11,8 @@
 
 namespace PHPSTORM_META
 {
-    expectedArguments(\League\CommonMark\Context::setEncoding(), 0, 'UTF-8', 'ASCII', 'ISO-8859-1');
-    expectedArguments(\League\CommonMark\Cursor::__construct(), 1, 'UTF-8', 'ASCII', 'ISO-8859-1');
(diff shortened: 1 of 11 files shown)


GrahamCampbell:
Seems like a good move to me. 👍

PR #366, opened 2019-05-25
"Final"ize all the things

colinodell:
Basically anything that is now completed covered by one or more interfaces (no other `public` or `protected` methods) has been marked `final`.  See https://ocramius.github.io/blog/when-to-declare-classes-final/ for some rationale behind this.


samwilson:
Is this a firm policy? I'm [trying to extend](https://github.com/thephpleague/commonmark/discussions/1076) the SmartPunct classes for LaTeX output but can't. Would you be willing to look at a change to those classes to remove `final`? (I'm happy to make a PR.) Otherwise I'm stuck [duplicating the code](https://github.com/samwilson/commonmark-latex/pull/14).

PR #368, opened 2019-05-27
Standardize delimiter processing node types

colinodell:
This fixes a couple of type issues found in 1.0.0-beta1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
 
 ## [Unreleased][unreleased]
 
+### Changed
+
+ - `DelimiterProcessorInterface::process()` will accept any type of `AbstractStringContainer` now, not just `Text` nodes
+ - The `Delimiter` constructor, `getInlineNode()`, and `setInlineNode()` no longer accept generic `Node` elements - only `AbstractStringContainer`s
+
 ## [1.0.0-beta1] - 2019-05-26
 
 ### Added
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,12 @@
 
 **Note:** This file has been deprecated.  Future upgrade instructions can be found on our website: <https://commonmark.thephpleague.com/releases>
 
+## UNRELEASED
+
+### Delimiter Processing
+
+We've standardized the new delimiter processing to work with `AbstractStringContainer` nodes.  Previously in 1.0.0-beta1, some parts (incorrectly) accepted any type of `Node`, while others forced you to use `Text` elements instead.
+
 ## 1.0.0-beta1
 
 ### Text Encoding
--- a/src/Delimiter/Delimiter.php
+++ b/src/Delimiter/Delimiter.php
@@ -14,7 +14,7 @@
 
 namespace League\CommonMark\Delimiter;
 
-use League\CommonMark\Node\Node;
+use League\CommonMark\Inline\Element\AbstractStringContainer;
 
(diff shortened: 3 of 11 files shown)


GrahamCampbell:
Could delimiters be made final?

colinodell:
Maybe.  I'm wondering if anyone ever has the need to extend this class to store additional information about their delimiter.  Probably not, so making this `final` should be okay.  We could extract a `DelimiterInterface` just in case :)

colinodell:
On second thought, I think I'd like to avoid the interface for now.  Delimiters seem more likely than other aspects to change in a future spec version, so I'd rather have the flexibility to easily add a method to a `final Delimiter` class vs having an interface I can't change as easily in a future version.

And if we ever decide to bump the minimum PHP version to 7.2 in a future 1.x minor release* then we do have the option to leverage contravariant typing to insert an interface later.

_* I know some people have different opinions on this, but [I side with the Doctrine team here](https://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html) - updating dependency versions and taking advantage of them without breaking your public API [is not a BC break](https://github.com/semver/semver/blob/df7bd79bda7d7fe6da20d0724fe0111678cbaa8f/semver.md#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api)._

GrahamCampbell:
I agree with all that you just said. 👍

colinodell:
Well upon further review, we may have a _slight_ issue - now that `Delimiter` is `final` we can't mock it in extension tests like this: https://github.com/thephpleague/commonmark-ext-smartpunct/blob/11b414730dc13603379dff548b74cb4b38fc37cc/tests/QuoteProcessorTest.php#L29

Sure, you could just instantiate a real one, but the constructor isn't the friendliest to use  :disappointed: 

So do we make testing easier for people or do we keep things as-is to mitigate the need to bump major versions when things change? :thinking: I'm thinking we should probably just bite the bullet and add the interface now and deal with things when they happen.  Thoughts?

colinodell:
Meh, let's just add the interface now.

The more I think about it, the more I feel like any change to delimiter handling is going to change much more than just this one class and require a major version bump anyway, so we should just do things "the right away" and provide the best experience in this 1.x major version.

colinodell:
Implemented in #371

GrahamCampbell:
Oh, hek. Do we have to mock it?

GrahamCampbell:
Feels wrong for a data class to have an interface.

GrahamCampbell:
Makes me think of https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition. :trollface:

colinodell:
> Feels wrong for a data class to have an interface.

Yeah it does... but the Java CommonMark library (which I drew heavy inspiration from since 0.19) uses an interface too: https://github.com/atlassian/commonmark-java/blob/6101e8a6cf43bcb6dde29ca3ee7be1e07e2c5ed2/commonmark/src/main/java/org/commonmark/internal/Delimiter.java

Furthermore, although I haven't seen it myself, it's possible that some people might want to store additional data on the delimiter during the initial parsing step and then refer to that data when it's time to look at all of the delimiters we've found.  The only sane way to do that would be extending `Delimiter` (which we no longer allow because we made this final) or to implement the interface the underlying engine needs to handle all of the other things it does with delimiters.

So yeah, we'll go with it.  Even if this does make the code look more like Java :upside_down_face: I think the reasons are just barely convincing enough that it isn't a totally bad idea haha

GrahamCampbell:
CommonMark EE 🎢

PR #372, opened 2019-06-04
Implement event dispatching

colinodell:
Instead of saving this for 2.0, I think it would be very helpful if 1.0 had a generic way to add future "hooks" into the library without needing to break backward compatibility - a basic event dispatcher implementation like this should do the trick.

I considering using PSR-14, but that would require bumping the minimum PHP version to 7.2, which I'm not interested in doing at this time - many people are using this on PHP 7.1 and I want to ensure they have a stable release too.  I also considered pulling in symfony/event-dispatcher but we only need a basic implementation for now.  We can re-evaluate these choices in 2.0.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -18,6 +18,7 @@ use League\CommonMark\Block\Parser\BlockParserInterface;
 use League\CommonMark\Block\Renderer\BlockRendererInterface;
 use League\CommonMark\Delimiter\Processor\DelimiterProcessorCollection;
 use League\CommonMark\Delimiter\Processor\DelimiterProcessorInterface;
+use League\CommonMark\Event\AbstractEvent;
 use League\CommonMark\Extension\CommonMarkCoreExtension;
 use League\CommonMark\Extension\ExtensionInterface;
 use League\CommonMark\Inline\Parser\InlineParserInterface;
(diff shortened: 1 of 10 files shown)


src/Environment.php
```diff
@@ -380,6 +386,42 @@ public function getInlineParserCharacterRegex(): string
...
+        $this->initializeExtensions();
+
+        $type = get_class($event);
+
+        foreach ($this->listeners[$type] ?? [] as $listener) {
+            if ($event->isPropagationStopped()) {
+                break;
+            }
+
+            call_user_func($listener, $event);
```
GrahamCampbell: Can we not just do `$listener($event);`? It should be faster to execute too.

src/EnvironmentInterface.php
```diff
@@ -75,4 +76,11 @@ public function getInlineRenderersForClass(string $inlineClass): iterable;
      * @return string
      */
     public function getInlineParserCharacterRegex(): string;
+
+    /**
+     * @param AbstractEvent $event
+     *
+     * @return AbstractEvent The processed event
+     */
+    public function dispatch(AbstractEvent $event): AbstractEvent;
```
GrahamCampbell: Why return the event vs void?

src/Environment.php
colinodell: Good call!

src/EnvironmentInterface.php
colinodell: I was thinking for ease of chaining:

```php
$foo = $environment->dispatch($event)->getFoo();
```

But I suppose that's unnecessary :)

PR #378, opened 2019-06-29
Fix max_nesting_level example value type on docs/security page

NicolasCARPi:
max_nesting_level is either INF (default, float), or an integer, it
doesn't make sense to use a string in the documentation.

------------------

Hello,

Spotted this tiny error in the documentation ;)

Cheers,
~Nico


colinodell:
Thank you, good catch!

PR #390, opened 2019-12-03
Unescaped pipe sign hid the errors

szepeviktor:
...of the last 8 months.

Please see [CI output](https://travis-ci.org/thephpleague/commonmark/jobs/620049465).

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -3,7 +3,7 @@ parameters:
   paths:
     - src
   ignoreErrors:
-    - "#should return string but returns League\\CommonMark\\HtmlElement|string#"
-    - "#expects int, float|int given#"
+    - '#should return string but returns League\\CommonMark\\HtmlElement\|string\.$#'
+    - '#expects int, float\|int given\.$#'
     - path: src/Delimiter/DelimiterStack.php
-      message: "#Cannot call method .+ on League\\CommonMark\\Delimiter\\Delimiter|null#"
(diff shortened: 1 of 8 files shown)


szepeviktor:
>   Ignored error pattern #should return string but returns League\\CommonMark\\HtmlElement\|string\.$# was not matched in reported errors.                                          
>  Ignored error pattern #Cannot call method .+ on League\\CommonMark\\Delimiter\\Delimiter\|null\.$# in path src/Delimiter/DelimiterStack.php was not matched in reported errors.  

These two seems to be resolved.

colinodell:
Yikes! Thank you for catching this.

szepeviktor:
Please consider fixing the revealed problem instead of adding them to ignore list.

colinodell:
I've resolved most of the issues that were found.  I've also chosen to ignore some errors, particularly ones about functions not expecting null or false values in cases where I know a `null` or `false` will never be passed, either due to the underlying logic (which phpstan doesn't understand) or when using `preg_` functions and I'm 100% certain there's no error with the regex.

szepeviktor:
All green 🍏

szepeviktor:
> Please enter RIPS-API URL (default: https://api-3.ripstech.com): 

@colinodell this env. var. may be missing

colinodell:
Hmm... it's definitely configured within Travis: 
image  So I'm not sure why it's asking for one here.

GrahamCampbell:
It's because Travis doesn't let PRs see secrets, so someone can't just PR an edit that prints the secret env variables.

szepeviktor:
Thank you.

PR #391, opened 2019-12-03
Test on PHP 7.4 stable

GrahamCampbell:
NB `sudo` has defaulted to `false` for a while (even though they initially only defaulted it for new repos).


szepeviktor:
@GrahamCampbell `composer self-update` is already in Travis images.

PR #409, opened 2020-02-04
Github-Flavored Markdown

colinodell:
**This pull request adds complete GFM support to this library.**

It will be an optional extension bundled with this library that anyone can easily use without needing to manually install and register external extensions.

## Background

Historically, this library has tried to avoid including any functionality not present in the CommonMark spec.  This included all features of Github-Flavored Markdown, such as tables.  Due to popular demand, we did eventually release those as standalone packages under the `league/*` Packagist namespace.  The idea was this would ensure proper separation and allow us to gradually develop and refactor those features over time.

However, the burden to maintain those various extensions as completely separate packages has been difficult.  Changes to the core parser were not automatically tested against those other extensions either.  And perhaps the biggest problem was not having a single extension unifying all the GFM-related extensions together.

## Bundled Extensions

**The following extensions will be added to this library once 1.3.0 is released:**

- **`GithubFlavoredMarkdownExtension`, a brand new extension which includes all of the following:**
  - `AutolinkExtension` (from `league/commonmark-ext-autolink`)
  - `DisallowedRawHtmlExtension` (new extension)
  - `StrikethroughExtension` (from `league/commonmark-ext-strikethrough`)
  - `TableExtension` (from `league/commonmark-ext-table`)
  - `TaskListExtension` (from `league/commonmark-ext-task-list`)

Two other extensions are also being included in this PR, even though they're not part of the GFM spec:

- **`SmartPunctExtension`** (from `league/commonmark-ext-smartpunct`)
- **`InlinesOnlyExtension`** (from `league/commonmark-ext-inlines-only`)

Those two extensions are not part of the GFM spec, but they do provide useful functionality based on the core library and so including them here meets the goal of reducing the maintenance burden.

## Deprecation and Migration

The following packages will be deprecated once `league/commonmark` 1.3.0 is released:

 - `league/commonmark-extras` (basically was a meta-package for most of the GFM extensions below)
- `league/commonmark-ext-autolink`
- `league/commonmark-ext-smartpunct`
- `league/commonmark-ext-strikethrough`
- `league/commonmark-ext-table`
- `league/commonmark-ext-task-list`
- `league/commonmark-ext-smartpunct`
- `league/commonmark-ext-inlines-only`

Users of those packages may continue to use them, but support for issues and future compatibility will not be provided.  **Users will therefore be strongly encouraged to switch to the built-in GFM support provided in 1.3.0**

Switching should be relatively painless for most people - the general gist is:

 - Upgrade to `league/commonmark` 1.3
 - Replace any references to the `League\CommonMark\Ext\` namespace in your code to `League\CommonMark\Extension\`
 - Remove the deprecated packages from your `composer.json`

## Extension Changes

For the most part, all of the extensions were imported as-is into this codebase with a single code-level change to the namespace:

```diff
-namespace League\CommonMark\Ext\{extension-name-here};
+namespace League\CommonMark\Extension\{extension-name-here};
```

A small number of other changes were made to specific extensions to align their behavior with the GFM spec, which may affect how some Markdown is parsed:

- **Tables**
  - Table captions are no longer supported (they are not part of the GFM spec)
  - Tables with a mismatched number of columns now behave properly
  - Tables use the 'align' attribute for alignment instead of hard-coded CSS
  - Escaped pipe characters are now handled properly
- **Task Lists**
  - The rendered `<input>` elements no longer look like `<input />` (with that closing slash)


GrahamCampbell:
Doesn't GFM have emojis?

colinodell:
> Doesn't GFM have emojis?

Great question!  [Emojis are actually implemented outside of their Markdown stack](https://github.com/github/cmark-gfm/issues/75#issuecomment-355440087) and are not part of [the GFM spec extensions](https://github.github.com/gfm/).  Based on that, I think it makes sense to keep the `:emoji:` syntax outside of the GFM extension unless the GFM spec includes them one day.

GrahamCampbell:
Ah, interesting. I wonder what other features they have that are outside of the spec, other than the at mentions, and issue links?

src/Block/Renderer/ListItemRenderer.php
```diff
@@ -16,7 +16,9 @@
 
 use League\CommonMark\Block\Element\AbstractBlock;
 use League\CommonMark\Block\Element\ListItem;
+use League\CommonMark\Block\Element\Paragraph;
 use League\CommonMark\ElementRendererInterface;
+use League\CommonMark\Extension\TaskList\TaskListItemMarker;
```
GrahamCampbell: Kinda an unfortunate cyclic dependency here.

colinodell: Yeah, especially since the only purpose is to ensure the output exactly matches the GFM spec not adding a newline :upside_down_face:  Oh well

glensc:
@colinodell the 1.3.0 is released, could you put the migration things to docs as well? I was expecting something about it in 1.3.0 release notes:
- https://github.com/thephpleague/commonmark/blob/1.3.0/CHANGELOG.md
- https://github.com/thephpleague/commonmark/releases/tag/1.3.0

colinodell:
I've added information to the tag's release notes to help point people in the right direction.  I also plan to release one more minor/bugfix release of each individual package with a Composer `suggests` line pointing people to the same instructions to help spread the word.

j4kim:
Is it possible to use `GithubFlavoredMarkdownConverter` but allow HTML raw inputs ?

docs/1.0/customization/extensions.md
```diff
@@ -34,3 +35,50 @@ $environment->addExtension(new EmojiExtension();
...
 echo $converter->convertToHtml('Hello! :wave:');
 ```
+
+## Included Extensions
+
+Starting in v1.3, this library includes several extensions to support Github-Flavored Markdown.  You can manually add the GFM extension to your environment like this:
+
+```php
+$environment = Environment::createCommonMarkExtension();
+$environment->addExtension(new GithubFlavoredMarkdownExtension();
```
j4kim: Missing a `)`:
```php
$environment->addExtension(new GithubFlavoredMarkdownExtension());
```

docs/1.0/customization/extensions.md
```diff
@@ -34,3 +35,50 @@ $environment->addExtension(new EmojiExtension();
...
+
+$converter = new CommonMarkConverter([], $environment);
+echo $converter->convertToHtml('Hello GFM!');
+
+```
+
+Or, if you only want a subset of GFM extensions, you can add them individually like this instead:
+
+```php
+$environment = Environment::createCommonMarkExtension();
```
j4kim: `createCommonMarkEnvironment` ?

docs/1.0/customization/extensions.md
```diff
@@ -34,3 +35,50 @@ $environment->addExtension(new EmojiExtension();
 $converter = new CommonMarkConverter([], $environment);
 echo $converter->convertToHtml('Hello! :wave:');
 ```
+
+## Included Extensions
+
+Starting in v1.3, this library includes several extensions to support Github-Flavored Markdown.  You can manually add the GFM extension to your environment like this:
+
+```php
+$environment = Environment::createCommonMarkExtension();
```
j4kim: This method does not seem to exist...
`createCommonMarkEnvironment` ?

docs/1.0/customization/extensions.md
```diff
@@ -34,3 +35,50 @@ $environment->addExtension(new EmojiExtension();
...
+$environment->addExtension(new GithubFlavoredMarkdownExtension();
+
+$converter = new CommonMarkConverter([], $environment);
+echo $converter->convertToHtml('Hello GFM!');
+
+```
+
+Or, if you only want a subset of GFM extensions, you can add them individually like this instead:
+
+```php
```
j4kim: Namespaces may be useful
```php
use League\CommonMark\Environment;
use League\CommonMark\Extension\TaskList\TaskListExtension;
use League\CommonMark\Extension\Autolink\AutolinkExtension;
// etc...
```

docs/1.0/customization/extensions.md
colinodell: Good catch!  This has been fixed in 34829b26

docs/1.0/customization/extensions.md
colinodell: Yep, thanks for catching that! Fixed in 34829b26

docs/1.0/customization/extensions.md
colinodell: Fixed in 34829b26

docs/1.0/customization/extensions.md
colinodell: Added in 34829b26

PR #420, opened 2020-04-04
Implement new heading permalink extension

colinodell:
New extension which automatically adds anchor links to headings, similar to Github:

image

**Documentation:** https://github.com/thephpleague/commonmark/blob/heading-link-extension/docs/1.3/extensions/heading-permalinks.md


colinodell:
I've just force-pushed a few changes:

 - Added some additional configuration options
 - Renamed things (because naming is hard)
 - Whipped up some documentation: https://github.com/thephpleague/commonmark/blob/heading-link-extension/docs/1.3/extensions/heading-permalinks.md

Any feedback would be appreciated!

glensc:
both links here point to 1.3, but I think they got released as 1.4 and there's no redirection, so the links are borken.

Issue #359, opened 2019-04-20
[RFC] Event dispatcher


colinodell:
Let's consider adding support for an (optional) event dispatcher which would allow people to hook into certain `private` methods.  For example, somebody may want to pre-populate references on the `Document` before we begin parsing.  This is not currently possible as we create that node and begin parsing within a `private` method.  Adding support for something like PSR-14 would make it easy for others to hook into this type of logic.

colinodell:
Marking this as a possible 2.0 feature - there's no demand for this yet and I'd like to better see how PSR-14 is accepted by the community before committing to a certain way of doing things.

colinodell:
Just kidding, let's target 1.0 via #372

colinodell:
Implemented via #372

PR #427, opened 2020-04-07
Pre-parsing events

neochief:
I have a site with a huge amount of markdown content that uses inline references extensively. Prior to 0.19 I was able to implement reference map caching by overriding DocParser, but now it's final, so it's no longer possible.

It seems that the ability to pre-fill the reference map was envisioned from the very beginning (https://github.com/thephpleague/commonmark/issues/359), so this pull request is a basic attempt to make it happen with events.

### DIFF
--- a/docs/1.4/customization/event-dispatcher.md
+++ b/docs/1.4/customization/event-dispatcher.md
@@ -67,6 +67,10 @@ Listeners may call any method on the event to get more information about the eve
 
 This library supports the following default events which you can register listeners for:
 
+### `League\CommonMark\Event\DocumentPreParsedEvent`
+
+This event is dispatched just before any processing is done. It can be used to pre-populate reference map of a document before any processing is performed.
+
 ### `League\CommonMark\Event\DocumentParsedEvent`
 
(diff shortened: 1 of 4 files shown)


colinodell:
This looks great!  And it's a perfect use of the new event system :)

However, I do need to complete https://github.com/thephpleague/commonmark/issues/428 first so that we can document new features like this one (and #420) in the proper location.  I'll plan to tackle that this week so we can rebase this and target the 1.4 docs instead.

neochief:
Awesome, thanks!

docs/1.3/customization/event-dispatcher.md
```diff
@@ -69,6 +69,10 @@ Listeners may call any method on the event to get more information about the eve
 
 This library supports the following default events which you can register listeners for:
 
+### `League\CommonMark\Event\DocumentPreParsedEvent`
+
+This event is dispatched just before any processing is done. It can be used to pre-populate reference map of a document before any processing is performed.
+
```
colinodell, line 75: I've added the `docs/1.4` folder.  Would you mind rebasing this PR onto `master` and apply this change to `1.4` instead?

neochief: Done!

PR #437, opened 2020-04-12
Deprecations for 1.4

colinodell:


bin/commonmark
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env php
 <?php
 
+fwrite(STDERR, "[DEPRECATED] This command has been deprecated and will be removed in v2.0\n\n");
```
GrahamCampbell, line 4: Why not raise a silenced deprecation error?

colinodell: I generally prefer that (and should probably add that to the other deprecations here, now that I think of it).  But in this particular case, there's no logger hooked up to this script which will detect them, so I don't think a silenced deprecation error would be noticed.  That being said, I suppose an unsilenced one could work.

colinodell: This has been replaced with a non-silenced deprecation error (and force-pushed).

bin/commonmark
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env php
 <?php
 
+trigger_error('The "bin/commonmark" command has been deprecated since league/commonmark 1.4', E_USER_DEPRECATED);
```
GrahamCampbell, line 4: Symfony does these silenced. https://github.com/symfony/debug/blob/v4.4.7/ErrorHandler.php#L26

colinodell: Right, but that only "works" if you have a custom error handler hooked up which can detect these:

> Without the @-silencing operator, users would need to opt-out from deprecation notices. Silencing swaps this behavior and allows users to opt-in when they are ready to cope with them (**by adding a custom error handler** like the one used by the Web Debug Toolbar or by the PHPUnit bridge).

(from https://symfony.com/doc/current/contributing/code/conventions.html#removing-deprecated-code - emphasis added)

This particular script is a standalone "executable" which does not provide an opportunity for `ErrorHandler::register()` to be called beforehand.  (Symfony does this by calling `Debug::enable()` in the `public/index.php` or `bin/console` files during the bootstrap process)

colinodell:
The `rips` job is failing due to rate-limiting - all other checks are passing.

PR #439, opened 2020-04-13
Add MarkdownConverterInterface; deprecate Converter and ConverterInterface

### LINKED ISSUE #438: Fully deprecate the Converter class
#437 deprecated just the constructor, but upon further review, we can deprecate the entire class and just use `CommonMarkConverter` in 2.0.

colinodell:
Per #438, the `Converter` class doesn't provide any value - it's an unnecessary level of abstraction that nobody really uses.

Deprecating `ConverterInterface` in favor of `MarkdownConverterInterface` is also being done here to make its purpose more clear.  And if we decide to implement HTML-to-Markdown conversion in this library in the future, we can do so by simply implementing a new `HtmlConverterInterface` on the `CommonMarkConverter` in a minor version release (vs adding a new method on the existing interface which would require a major version bump).

Resolves #438

### DIFF
--- a/src/CommonMarkConverter.php
+++ b/src/CommonMarkConverter.php
@@ -47,8 +47,7 @@ class CommonMarkConverter extends Converter
 
         $this->environment = $environment;
 
-        $this->docParser = new DocParser($environment);
-        $this->htmlRenderer = new HtmlRenderer($environment);
+        parent::__construct(new DocParser($environment), new HtmlRenderer($environment));
     }
 
     /**
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -13,8 +13,10 @@ namespace League\CommonMark;
 
 /**
  * Converts CommonMark-compatible Markdown to HTML.
+ *
+ * @deprecated This class is deprecated since league/commonmark 1.4, use CommonMarkConverter instead.
  */
-class Converter implements ConverterInterface
+class Converter implements MarkdownConverterInterface
 {
     /**
      * The document parser instance.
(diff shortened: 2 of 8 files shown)


src/CommonMarkConverter.php
```diff
@@ -47,8 +47,7 @@ public function __construct(array $config = [], EnvironmentInterface $environmen
 
         $this->environment = $environment;
 
-        $this->docParser = new DocParser($environment);
-        $this->htmlRenderer = new HtmlRenderer($environment);
+        parent::__construct(new DocParser($environment), new HtmlRenderer($environment));
```
colinodell, line 50: Undoes 66e928ece5f2f735446c6cfcb08f97465954c0af

src/Converter.php
```diff
@@ -35,12 +37,12 @@ class Converter implements ConverterInterface
...
      * @param ElementRendererInterface $htmlRenderer
-     *
-     * @deprecated Instantiating a Converter class with a DocParserInterface and ElementRendererInterface is deprecated since league/commonmark 1.4. In 2.0, this constructor will require a configuration array and EnvironmentInterface.
      */
     public function __construct(DocParserInterface $docParser, ElementRendererInterface $htmlRenderer)
     {
-        @trigger_error('Instantiating a "Converter" class with a DocParserInterface and ElementRendererInterface is deprecated since league/commonmark 1.4. In 2.0, this constructor will require a configuration array and EnvironmentInterface.', E_USER_DEPRECATED);
+        if (!($this instanceof CommonMarkConverter)) {
+            @trigger_error(sprintf('The %s class is deprecated since league/commonmark 1.4, use %s instead.', self::class, CommonMarkConverter::class), E_USER_DEPRECATED);
+        }
```
colinodell, line 45: We don't want to throw this error if somebody is using CommonMarkConverter directly or as a base class.

Issue #441, opened 2020-04-13
Feature Request: Table of Contents Extension


colinodell:
I've begun working on an extension that can generate a Table of Contents.  It relies on the new Heading Permalink extension, which already identifies headers and makes them linkable.  This new extension will scan for those links and build a `TableOfContents` element accordingly.

Likely configuration settings will include:

 - **How to "normalize" the ToC:**
    - Show in a tree-like structure based on the heading levels
    - Same as above, but use relative nesting levels instead of exact levels (I can't explain this very well, but I know what this means and will provide docs for it)
    - "Flatten" them into a single-level list
 - **Where to insert the ToC:**
    - Very top of the document
    - Just before the first heading
 - **Minimum and maximum heading levels to include**
    - Maybe you only want links for `h1` and `h2` elements, maybe you want them all

GrahamCampbell:
<img src="https://cultofthepartyparrot.com/parrots/hd/parrot.gif" data-canonical-src="https://cultofthepartyparrot.com/parrots/hd/parrot.gif" width="32" height="32" />

colinodell:
The implementation I'm working on now does indeed use a TOC element.  I think we should allow the user to select one of those two options, as they're probably the most common, but anyone could register a custom event listener that runs after ours to locate that element and move it somewhere else.

theREALdebater:
I suggest you've got slightly the wrong idea. 
You don't want a ToC feature in the tool that converts MD to HTML. Instead you want a pre-processing tool---either a separate tool or an extension to an existing one---that sees a header named `Contents` and inserts the table of contents, as an MD bullet-list, into the MD. If there is already text following the `Contents` header, it is first deleted, so you can re-run the pre-processor whenever there have been changes to the MD file to regenerate the ToC. 
The reason why is because, to my mind, you want the ToC in the MD as well as in any transformed format (HTML, PDF, or whatever).

colinodell:
I appreciate the feedback.  Right now, this works by inserting a `TableOfContents` node into the parsed Abstract Syntax Tree.  This `TableOfContents` node can be read, moved, deleted, etc. fairly easily.  All of this happens prior to any type of rendering, thus allowing you to implement that type of functionality.

However, I will admit that the current implementation tightly-couples the generation of the TOC with inserting it into the AST.  Perhaps we could decouple them into two classes: one responsible for generating the TOC and another responsible for adding it to the AST.  This would provide more flexibility to customize the TOC behavior to your liking.

PR #445, opened 2020-04-17
ListBlock constant changes

colinodell:
These minor adjustments will better-align the constants with the CommonMark spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
  - Added new [Heading Permalink extension](https://commonmark.thephpleague.com/extensions/heading-permalinks/) (#420)
  - Added new `MarkdownConverterInterface` as a long-term replacement for `ConverterInterface` (#439)
  - Added new `DocumentPreParsedEvent` event (#427, #359, #399)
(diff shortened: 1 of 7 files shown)


colinodell:
Because this includes a change to `const` values, I am [following Symfony's guidelines](https://symfony.com/doc/current/contributing/code/bc.html#id65) here:

> Should be avoided. When done, this change must be documented in the UPGRADE file.

It's less painful if we do this now.  The change is noted in both the changelog and our upgrading documentation.

> The value of a constant may only be changed when the constants aren't used in configuration (e.g. Yaml and XML files), as these do not support constants and have to hardcode the value. For instance, event name constants can't change the value without introducing a BC break. Additionally, if a constant will likely be used in objects that are serialized, the value of a constant should not be changed.

Why I feel this should not be considered a BC-break:

 - Within this library, the values of those constants are never used
 - Those constants are not currently tied to any configuration settings
 - This library does not currently support serialization
 - Anyone not hooking custom code into ListBlock functionality (most end users) will be unaffected
 - Any external, custom code using the original hard-coded, title-cased value won't experience any errors or exceptions at all; worst case, an ordered list _might_ render as an unordered one.
 - It's technically valid per semver

These constants _might_ become tied to configurations as part of #441 (slated for v1.4).  If not, we'll end up with two similar-but-different values serving similar purposes in our codebase which would be very confusing.  And, per those guidelines, we should avoid changing that after 1.4 anyway.

PR #453, opened 2020-04-20
Fixed branch alias

GrahamCampbell:
We can't have both a 1.4 branch and a master branch claiming that it is 1.4. ;)


colinodell:
Thanks for catching this!

PR #455, opened 2020-04-20
Test on PHP 7.2.5 lowest

GrahamCampbell:
Now we are testing the minimum version supported by the package, as well as with the lowest dependencies.


GrahamCampbell:
Hmm, travis doesn't seem to be building this. :/

PR #458, opened 2020-04-23
Document potential issues with additional HTML post-processing

glensc:
Add note that `inner_contents` might be filtered out by HTML post-processing or input filtering.

My use case was that post-processor removed the default `<svg>` node by htmlpurifier (with html5purifier):
- https://github.com/eventum/eventum/blob/v3.8.9/src/Markdown.php#L59

### DIFF
--- a/docs/1.4/security.md
+++ b/docs/1.4/security.md
@@ -87,3 +87,8 @@ echo $converter->convertToHtml($markdown);
 ~~~
 
 See the [configuration](/1.4/configuration/) section for more information.
+
+
+## Additional Filtering
+
+Although this library does offer these security features out-of-the-box, some users may opt to also run the HTML output through additional filtering layers (like HTMLPurifier).  If you do this, make sure you **thoroughly** test your additional post-processing steps and configure them to work properly with the types of HTML elements and attributes that converted Markdown might produce, otherwise, you may end up with weird behavior like missing images, broken links, mismatched HTML tags, etc.


docs/1.4/extensions/heading-permalinks.md
```diff
@@ -63,6 +63,8 @@ This controls the HTML you want to appear inside of the generated `<a>` tag.  Us
 
 By default, we provide an embedded [Octicon link SVG](https://octicons.github.com/icon/link/), but you can replace this with any custom HTML you wish.
 
+The inner contents value is candidate for HTML filtering, so it may become empty if you filter out `<svg>` tags in HTML.
```
colinodell, line 66: Any output of this library may be subject to filtering by external HTML filtering libraries, depending on which library you're using and how you have it configured.  While this particular item may have triggered filtering in your case, I don't think this "issue" is necessarily tied to this particular feature.

If we're going to add a warning/tip like this, I think it would be more appropriate to put a note towards the bottom of [the Security page](https://commonmark.thephpleague.com/1.4/security/) recommending that anyone who chooses to use additional HTML filtering should ensure their filtering is properly configured based on the Markdown features they are using.

glensc: Absolutely. I just wanted to write something down so this doesn't get lost. I was really confused for some moments that why example copied from documentation just did not work. In hope that someone else would be not spending time on same problem.

I don't get offended at all if you finish this up yourself :) (Allow edits from maintainers is enabled)

colinodell: Sure thing! I've force-pushed a change to put some similar documentation on the Security page instead.  I appreciate you bringing this up :)

Issue #232, opened 2016-02-03
Revisit and finalize the AST implementation


colinodell:
While the AST implementation is working great, there are some areas I'd like to revisit and see if changes could improve the implementation:
## Remove parsing-related methods from node types

Node types contain some methods which are only used for parsing Markdown input:
- `matchesNextLine()`
- `handleRemainingContents()`
- etc.

Can we somehow move these into their respective parsers?  Or alternatively, do we need to separate node types from node instances (similar to how the Symfony Form Component has "form types" and form objects)?
## Implement XML serialization

The reference parser supports rendering the AST as XML.  I would like to have the same functionality in league/commonmark.  This would also be a good test of using the AST for purposes other than rendering it as HTML  (converting MD -> XML, HTML -> MD, etc.)  However, we may need to implement the other changes in this list including the next one:

## Determine whether node types should be referenced by an (arbitrary) name instead of namespaced class

~When registering parsers and renderers, or checking the type of the node, we're currently relying on using PHP classes and namespaces.  For example, a paragraph is differentiated from a list by it's class of `League\CommonMark\Block\Element\Paragraph`.~

~We've also added the ability for renderers to handle subclasses which relies on the class hierarchy in PHP.~

~But in order to implement the XML serialization, we need to associate a name like `paragraph` to this element.~

~I'd therefore like to investigate whether we can use a system like Symfony's Form Component where each type declares its name (i.e. `paragraph`) and an optional parent type.  We'd then register each type in the environment like so:~

**Update: Let's keep using fully-namespaced classes and use a different approach to determine serialized XML node names**

```
$environment->addType('paragraph`, new League\CommonMark\Block\Type\Paragraph());
```

This `paragraph` name would then be used for things like:
- Registering parsers/renderers for each type
- Determining which renderer to use
- Naming XML elements in the AST
- Node type inheritance

However, [Symfony 3.0 removed the ability to use these "alias" names](https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md#form) and now require the fully-qualified class names.  I'd like to understand why this was done in order to determine whether this idea is good or not.
## Change the namespace from `Node` to `AST`

I'm not sure if I want to do this, but I do want to consider it.

---

I may add more to this list later.  Any thoughts or feedback are welcome!

yellow1912:
[...]

PR #462, opened 2020-05-03
Refactor parsing approach

### LINKED ISSUE #430: [2.0] Refactor Parsing Approach
In 1.x, functionality for parsing elements lives in two separate types of classes - both the parser and the element.  Node types contain some methods which are only used for parsing Markdown input:

 - `matchesNextLine()`
 - `handleRemainingContents()`
 - etc.

This logic should only live in one place, and definitely not in the element, which should **only** be a representation of the node in the AST.

Let's look into alternative implementations, such as Atlassian's Java fork, for some inspiration on how we can address this for v2.  For example, they basically keep track of which parser identified and created the node, allowing them to reference that later on.  This seems like a better approach to what I originally was considering in #232.

colinodell:
This PR changes how block parsing will work in 2.0.  In a nutshell, 1.x had parsing responsibilities split between the parser and the node. But nodes should be "dumb" and not know anything about how they are parsed - they should only know the bare minimum needed for rendering.

As a result, 2.x has delegated the parsing responsibilities to two different interfaces:

|  Responsibility                                           | Old Method (1.x)                           | New Method (2.0)                                               |
| --------------------------------------------------------- | ------------------------------------------ | -------------------------------------------------------------- |
| Identifying the start of a block                          | `BlockParserInterface::parse()`            | `BlockStartParserInterface::tryStart()`                        |
| Instantiating and configuring the new block               | `BlockParserInterface::parse()`            | `BlockContinueParserInterface::__construct()`                  |
| Determining if the block acts as a container              | `AbstractBlock::isContainer()`             | `BlockContinueParserInterface::isContainer()`                  |
| Determining if the block can have lazy continuation lines | `AbstractBlock::isCode()`                  | `BlockContinueParserInterface::canHaveLazyContinuationLines()` |
| Determining if the block can contain certain child blocks | `AbstractBlock::canContain()`              | `BlockContinueParserInterface::canContain()`                   |
| Determining if the block continues on the next line       | `AbstractBlock::matchesNextLine()`         | `BlockContinueParserInterface::tryContinue()`                  |
| Adding the next line to the block                         | `AbstractBlock::handleRemainingContents()` | `BlockContinueParserInterface::addLine()`                      |
| Finalizing the block and its contents                     | `AbstractBlock::finalize()`                | `BlockContinueParserInterface::closeBlock()`                   |

This approach also seems to be **35-40% faster than 1.x**, as it actively tracks the state of open block parsers instead of trying to recalculate it each time a new line is added.

-----

Resolves #430 and #232


colinodell:
Travis CI is failing due to passing around types that are too wide.  Implementing #463 would resolve these issues, so I may merge this as-is and address those issues in a subsequent PR.

PR #480, opened 2020-05-16
Allow configuration of rel attributes in ExternalLink extension

### LINKED ISSUE #476: Feature Request: Add rel attribute support on ExternalLinks
**Description**  
Currently, the `rel` attribute is hardcoded to `noopener noreferrer` with no support for `nofollow` at all.

Based on https://git.drupalcode.org/project/markdown/-/blob/8.x-2.x/src/Plugin/Markdown/CommonMark/Extension/ExternalLinkExtension.php#L119-145:

Add new config options:

- `external_link/nofollow`
- `external_link/noopener`
- `external_link/noreferrer`

That can be set to one of the following values:

- ` ` - No links (empty string)
- `all` - All links
- `external` - External links only
- `internal` - Internal links only

Then the `rel` attribute can be determined provided these settings:
https://git.drupalcode.org/project/markdown/-/blob/8.x-2.x/src/Plugin/Markdown/CommonMark/Extension/ExternalLinkExtension.php#L211-227

colinodell:
Resolves #476


docs/1.5/extensions/external-links.md
```diff
@@ -30,9 +31,12 @@ $environment->addExtension(new ExternalLinkExtension());
 // Set your configuration
 $config = [
     'external_link' => [
-        'internal_hosts' => 'www.example.com', // Don't forget to set this!
+        'internal_hosts' => 'www.example.com', // TODO: Don't forget to set this!
         'open_in_new_window' => true,
         'html_class' => 'external-link',
+        'nofollow' => 'external',
```
markcarver, line 37: This should default to empty as it has serious SEO consequences and should be left as opt-in.

docs/1.5/extensions/external-links.md
```diff
@@ -30,9 +31,12 @@ $environment->addExtension(new ExternalLinkExtension());
...
+        'noopener' => 'all',
```
markcarver, line 38: This should default to `external` as internal links may need this capability.

docs/1.5/extensions/external-links.md
colinodell: These are just examples of possible values but I'll change them to be more-sane for anyone blindly copypasting.  Although I did actually have `nofollow` set to `external` by default, which I've changed via commit 3477aa21

docs/1.5/extensions/external-links.md
colinodell: It does, this just wasn't a good example to share :upside_down_face:

PR #482, opened 2020-05-17
Implement new MentionParser; deprecate the old one

### LINKED ISSUE #473: Feature Request: Allow InlineMentionParser to be a little more dynamic
**Description**  
Currently, in Drupal, we're having to reimplement this class to allow it to be a little more flexible so we can multi-purpose it. While we can certainly keep this up in our project, I figured these small additions might be something you'd be interested in porting upstream. If so, we wouldn't have to maintain our own version any longer.

Specifically, in two areas:

1. Dynamic symbol:
    - https://git.drupalcode.org/project/markdown/-/blob/1841fcb2d629ec40c8e83a3add861d2c871a36d2/src/Plugin/Markdown/CommonMark/Extension/InlineMentionParser.php#L45-46
    - https://git.drupalcode.org/project/markdown/-/blob/1841fcb2d629ec40c8e83a3add861d2c871a36d2/src/Plugin/Markdown/CommonMark/Extension/InlineMentionParser.php#L74
2. Dynamic link pattern handling (allowing a callable to be passed instead):
    - https://git.drupalcode.org/project/markdown/-/blob/1841fcb2d629ec40c8e83a3add861d2c871a36d2/src/Plugin/Markdown/CommonMark/Extension/InlineMentionParser.php#L108-126

This allows us to, more easily, provide multiple ways to "map" an `@` inline mention (entity, github, twitter and custom) as well as `#` (entity, custom):

- https://git.drupalcode.org/project/markdown/-/blob/1841fcb2d629ec40c8e83a3add861d2c871a36d2/src/Plugin/Markdown/CommonMark/Extension/AutolinkExtension.php#L298-327
- https://git.drupalcode.org/project/markdown/-/blob/1841fcb2d629ec40c8e83a3add861d2c871a36d2/src/Plugin/Markdown/CommonMark/Extension/AutolinkExtension.php#L201-250
- https://git.drupalcode.org/project/markdown/-/blob/1841fcb2d629ec40c8e83a3add861d2c871a36d2/src/Plugin/Markdown/CommonMark/Extension/AutolinkExtension.php#L252-289

**Example**  
Screen Shot 2020-05-12 at 7 34 18 AM

colinodell:
Closes #473

### DIFF
--- a/src/Extension/Mention/MentionParser.php
+++ b/src/Extension/Mention/MentionParser.php
@@ -11,8 +11,9 @@
 
 namespace League\CommonMark\Extension\Mention;
 
-use League\CommonMark\Inline\Element\AbstractInline;
-use League\CommonMark\Inline\Element\Link;
+use League\CommonMark\Extension\Mention\LinkGenerator\CallbackLinkGenerator;
+use League\CommonMark\Extension\Mention\LinkGenerator\MentionLinkGeneratorInterface;
+use League\CommonMark\Extension\Mention\LinkGenerator\StringTemplateLinkGenerator;
 use League\CommonMark\Inline\Parser\InlineParserInterface;
 use League\CommonMark\InlineParserContext;
 
@@ -24,14 +25,14 @@ final class MentionParser implements InlineParserInterface
     /** @var string */
     private $mentionRegex;
 
-    /** @var callable(string, string): string|null */
-    private $urlGenerator;
+    /** @var MentionLinkGeneratorInterface */
+    private $linkGenerator;
 
-    public function __construct(string $symbol, string $mentionRegex, callable $urlGenerator)
+    public function __construct(string $symbol, string $mentionRegex, MentionLinkGeneratorInterface $linkGenerator)
     {
         $this->symbol = $symbol;
         $this->mentionRegex = $mentionRegex;
-        $this->urlGenerator = $urlGenerator;
+        $this->linkGenerator = $linkGenerator;
     }
 
     public function getCharacters(): array
@@ -65,35 +66,27 @@ final class MentionParser implements InlineParserInterface
             return false;
         }
 
-        $url = \call_user_func($this->urlGenerator, $handle, $this->symbol);
-        if ($url === null) {
+        $link = $this->linkGenerator->generateLink($this->symbol, $handle);
+
+        if ($link === null) {
             $cursor->restoreState($previousState);
 
             return false;
         }
 
-        if ($url instanceof AbstractInline) {
-            $inlineContext->getContainer()->appendChild($url);
-
-            return true;
-        }
-
-        if (!\is_string($url)) {
-            throw new \RuntimeException('Invalid URL - URL generator must return a string, AbstractInline, or null');
-        }
-
-        $inlineContext->getContainer()->appendChild(new Link($url, $this->symbol . $handle));
+        $inlineContext->getContainer()->appendChild($link);
 
         return true;
     }
 
     public static function createWithStringTemplate(string $symbol, string $mentionRegex, string $urlTemplate): MentionParser
     {
-        $urlGenerator = function (string $mention) use ($urlTemplate): string {
-            return \sprintf($urlTemplate, $mention);
-        };
+        return new self($symbol, $mentionRegex, new StringTemplateLinkGenerator($urlTemplate));
+    }
 
-        return new self($symbol, $mentionRegex, $urlGenerator);
+    public static function createWithCallback(string $symbol, string $mentionRegex, callable $callback): MentionParser
+    {
+        return new self($symbol, $mentionRegex, new CallbackLinkGenerator($callback));
     }
 
     public static function createTwitterHandleParser(): MentionParser
--- a/src/Extension/Mention/MentionParser.php
+++ b/src/Extension/Mention/MentionParser.php
@@ -1,0 +1,114 @@
+<?php
+
+/*
+ * This file is part of the league/commonmark package.
+ *
+ * (c) Colin O'Dell <colinodell@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace League\CommonMark\Extension\Mention;
+
+use League\CommonMark\Inline\Element\AbstractInline;
+use League\CommonMark\Inline\Element\Link;
+use League\CommonMark\Inline\Parser\InlineParserInterface;
+use League\CommonMark\InlineParserContext;
+
+final class MentionParser implements InlineParserInterface
+{
+    /** @var string */
+    private $symbol;
+
+    /** @var string */
+    private $mentionRegex;
+
+    /** @var callable(string, string): string|null */
+    private $urlGenerator;
+
+    public function __construct(string $symbol, string $mentionRegex, callable $urlGenerator)
+    {
+        $this->symbol = $symbol;
+        $this->mentionRegex = $mentionRegex;
+        $this->urlGenerator = $urlGenerator;
+    }
+
+    public function getCharacters(): array
+    {
+        return [$this->symbol];
+    }
+
+    public function parse(InlineParserContext $inlineContext): bool
+    {
+        $cursor = $inlineContext->getCursor();
+
+        // The symbol must not have any other characters immediately prior
+        $previousChar = $cursor->peek(-1);
+        if ($previousChar !== null && $previousChar !== ' ') {
+            // peek() doesn't modify the cursor, so no need to restore state first
+            return false;
+        }
+
+        // Save the cursor state in case we need to rewind and bail
+        $previousState = $cursor->saveState();
+
+        // Advance past the symbol to keep parsing simpler
+        $cursor->advance();
+
+        // Parse the handle
+        $handle = $cursor->match($this->mentionRegex);
+        if (empty($handle)) {
+            // Regex failed to match; this isn't a valid mention
+            $cursor->restoreState($previousState);
+
+            return false;
+        }
+
+        $url = \call_user_func($this->urlGenerator, $handle, $this->symbol);
+        if ($url === null) {
+            $cursor->restoreState($previousState);
+
+            return false;
+        }
+
+        if ($url instanceof AbstractInline) {
+            $inlineContext->getContainer()->appendChild($url);
+
+            return true;
+        }
+
+        if (!\is_string($url)) {
+            throw new \RuntimeException('Invalid URL - URL generator must return a string, AbstractInline, or null');
+        }
+
+        $inlineContext->getContainer()->appendChild(new Link($url, $this->symbol . $handle));
+
+        return true;
+    }
+
+    public static function createWithStringTemplate(string $symbol, string $mentionRegex, string $urlTemplate): MentionParser
+    {
+        $urlGenerator = function (string $mention) use ($urlTemplate): string {
+            return \sprintf($urlTemplate, $mention);
+        };
+
+        return new self($symbol, $mentionRegex, $urlGenerator);
+    }
+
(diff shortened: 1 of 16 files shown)


colinodell:
@markcarver How does this look?

It allows the same type of functionality as the original `InlineMentionParser`, but lets you plug in a custom `callable` to generate the URL (or even a custom `Link` with additional attributes set)

I've decided to keep `Mention` in the name because:

1. The primary use case is "mentioning" or "referencing" resources elsewhere in the site or the internet with a shortcut syntax
2. We can't use `Reference` because that's already a different concept in CommonMark
3. It's similar enough to the old name that people can still find it
4. `Token` feels a little too Drupal-specific
5. **Naming things is hard**

src/Extension/Mention/MentionParser.php
```diff
@@ -0,0 +1,114 @@
...
+            $inlineContext->getContainer()->appendChild($url);
+
+            return true;
+        }
+
+        if (!\is_string($url)) {
+            throw new \RuntimeException('Invalid URL - URL generator must return a string, AbstractInline, or null');
+        }
+
+        $inlineContext->getContainer()->appendChild(new Link($url, $this->symbol . $handle));
```
markcarver, line 85: Originally, I had stored `$this->symbol . $handle` as a `$label` variable that was passed by reference to the callable:

https://git.drupalcode.org/project/markdown/-/blob/1841fcb2d629ec40c8e83a3add861d2c871a36d2/src/Plugin/Markdown/CommonMark/Extension/InlineMentionParser.php#L109-112

This was, primarily, thinking ahead to avoid needing to know about/relying on an internal `League\CommonMark\Inline\Element\Link` class and allow the callable to modify/replace the label used in the link.

markcarver: All of this is fantastic BTW! I just think the power/benefit of abstracting this down should extend to not having to rely on returning a custom `Link` object which would require setting namespaces in stone.

An alternative, perhaps, would instead maybe create a `Mention` class that extends `Link` and has all the properties (symbol, handle, label, etc.) already set (and could be overridden in the callable, e.g. label). Then the object is passed as the sole parameter to the callable and expects the callable to invoke `setUrl` (which an exception could be thrown if it isn't set).

This would allow for a more "lazy" implementation of the callable where the type doesn't have to be explicitly specified (e.g. `function ($mention)`) and instead done so as a PHP comment that IDEs can use (i.e. `/* @var League\CommonMark\Inline\Element\Mention $mention */`).

This would prevent implementations from potentially breaking in the future just because of a namespace change.

markcarver: p.p.s. I'm not saying a callable still can't return whatever object that implements `AbstractInline` it wants to, I'm just thinking from the markdown module's perspective where I'd like to avoid using (creating) internal classes as much as possible. Getting handed an object as a parameter and then doing type checking using `class_exists` and `instanceof` would be far more preferable/future proof on our end.

colinodell: I appreciate the feedback!  I've made some changes which I think address your concerns.

I've basically added a layer of abstraction between the parsing and `Link` creation steps using a new [`MentionLinkGeneratorInterface`](https://github.com/thephpleague/commonmark/pull/482/commits/3b34e364688d004624c789f3cdc455bdbd2c4934#diff-9347709f4dfeeaf55db9cfceb13277c5).  We pass in the symbol and handle and the implementation returns a `Link` or `null`.

To help facilitate your particular use cases, I've also added [an implementation that works with callbacks](https://github.com/thephpleague/commonmark/pull/482/commits/3b34e364688d004624c789f3cdc455bdbd2c4934#diff-c0399fca28d895b50aaede9547cf555d) like the one in your example.

[The unit test](https://github.com/thephpleague/commonmark/pull/482/commits/3b34e364688d004624c789f3cdc455bdbd2c4934#diff-46f1024b40805b0af0e86f20521773b4) shows how it would look from your perspective.

Basically, you provide a callback which will receive three string parameters:

```php
$callback = function (string $handle, string &$label, string $symbol): ?string {
    // ...
};
```

The callable must return either:
  - A `string` containing the URL to use; or,
  - `null` if no `Link` should be generated

Note that the `$label` is passed by-reference - this allows you to also customize the label if desired.  The `$handle` and `$symbol` are not used by that "outer class" and are therefore not passed by reference (since modifying them would have no effect).

With this approach you don't need to know anything about the `Link` class - [the `CallbackLinkGenerator` class handles all of that for you](https://github.com/thephpleague/commonmark/pull/482/commits/3b34e364688d004624c789f3cdc455bdbd2c4934#diff-c0399fca28d895b50aaede9547cf555d).

I feel like this approach strikes the right balance with regards to separation of concerns, maintainability, and making it easy for external code (like the Drupal module) to customize things without knowing too much about the parser internals.  But what do you think?

markcarver: This will definitely work and I can live with that.

Just curious about your thoughts on creating a new `Mention` class that actually just extends `Link`: https://github.com/thephpleague/commonmark/pull/482#discussion_r426204191

That might be a better approach for long term sustainability (strongly typed objects over arbitrary parameters) and feature/future-proofing:

```php
<?php

$callback = function ($mention): ?string {
  if ($mention instanceof \League\CommonMark\Inline\Element\Mention) {
    // Follow what this object expects.
    $mention->setLabel($user->getFullName());
    $mention->setUrl($user->getUrl());
  }
  if ($mention instanceof \League\CommonMark\NewNamespaceFeatures\Mention) {
    // Follow what this object expects.
    $mention->setLabel($user->getFullName());
    $mention->setUrl($user->getUrl());
    $mention->setNewFeature();
  }
};

Also, this doesn't have the drawback of fataling if the class doesn't exist, like typecasting parameters would.

Comment (markcarver) [reply src/Extension/Mention/MentionParser.php]: @colinodell, sorry... updated example code

Comment (colinodell) [reply src/Extension/Mention/MentionParser.php]: From a high level, I like that approach, but I think there are two challenges with it.

First, `Link` (and its parent class `AbstractWebResource`) both require the `$url` to be a non-nullable string, so allowing a null value inside an instantiated object would technically break BC.

Additionally, the label is stored as a child `Text` node in the AST.  Adding a method like `setLabel()` would introduce a new pattern where descendants of `Node` are now manipulating the AST to add/remove child elements.  I'd like to keep descendants of `Node`  as dumb as possible - managing other nodes should be an external responsibility.

Technically v1.4 already has some of this - for example, the table extension we brought into this package has a `Table` element which keeps track of the table head and body elements: https://github.com/thephpleague/commonmark/blob/1.5/src/Extension/Table/Table.php  I've already started moving away from this "anti-pattern" in 2.x: https://github.com/thephpleague/commonmark/blob/master/src/Extension/Table/Table.php

That being said, I'm okay with having a _little_ bit of this in constructors that instantiate nodes, but once a specific node type is created it shouldn't directly manage anything else in the AST - parsers, processors, etc should be the only ones doing that.

Comment (markcarver) [reply src/Extension/Mention/MentionParser.php]: > First, `Link` (and its parent class `AbstractWebResource`) both require the `$url` to be a non-nullable string, so allowing a null value inside an instantiated object would technically break BC.

You don't have to call the parent constructor, nor does an extending class have to implement the same parameters of the parent constructor.



> Additionally, the label is stored as a child `Text` node in the AST. Adding a method like `setLabel()` would introduce a new pattern where descendants of `Node` are now manipulating the AST to add/remove child elements.

How are you manipulating something that doesn't exist? I would imagine that in the `setLabel()` method, you'd just convert it to the appropriate node type like you do when setting the initial label.

Comment (markcarver) [reply src/Extension/Mention/MentionParser.php]: Ok, well you merged, so it looks like this is what we're going with.

Comment (colinodell) [reply src/Extension/Mention/MentionParser.php]: > You don't have to call the parent constructor, nor does an extending class have to implement the same parameters of the parent constructor.

True, but if a `Mention` with a null `$url` is passed into something that wants an `AbstractWebResource` and that method calls `getUrl()` you'll get a `TypeError`: https://3v4l.org/1XuaT

We also can't change the return type of `Mention::getUrl()` to indicate the return value might be `null`: https://3v4l.org/LQ5BJ

And we can't change the return type of `AbstractWebResource::getUrl()` without it breaking BC.

So for those reasons we need to keep `$url` as a non-nullable in 1.x.

Comment (markcarver) [reply src/Extension/Mention/MentionParser.php]: Couldn't the URL just be an empty string to start? Then just checked in the MentionParser for a URL that isn't empty?

Comment (colinodell) [reply src/Extension/Mention/MentionParser.php]: Unfortunately I don't think there's a way to do that without resorting to workarounds like catching TypeErrors or adding a `hasUrl()` method.  This would "solve" the issue within `MentionParser` but would not for anyone else who might end up with a `Mention` instance.

If this library has a method saying we'll always return a `string` then I would like to keep that promise to all users :)

Comment (markcarver) [reply src/Extension/Mention/MentionParser.php]: Fair enough. So what if it doesn't extend `Link` and is just a new standalone/dedicated `Mention` object instead? I'm just worried that if we stick with arbitrary parameters (passed by reference), it will likely eventually bite us in the ass if it ever needs to be expanded upon.

Comment (markcarver) [reply src/Extension/Mention/MentionParser.php]: Then it could be a simple as

```php
$inlineContext->getContainer()->appendChild(new Link($mention->getUrl(), $mention->getLabel()));
```

colinodell: I did tinker with an implementation like that - it basically had 5 internal properties, the latter three of which could be modified:

 - symbol (string)
 - handle (string)
 - label (string; defaults to the concatenation of the two value above)
 - url (nullable string; default is null)
 - label (nullable string; default it null)

`MentionLinkGeneratorInterface` then looked like something this:

```
public function updateMention(Mention $mention): void
```

Creation of the `Link` was exactly what you had here.

However, I felt there were a two downsides to it:

1. It adds yet another layer of abstraction to build and maintain
2. Implementers don't have the ability to return subclasses of `Link` or add addition child nodes to it

I'll admit that point 1 is a very selfish argument, but let's table that for the moment.

But with regards to point 2, I could imagine somebody wanting to implement something custom that adds additional information to the AST.  For example, let's say somebody wanted to emulate what GitHub does when I put in `#473` into Markdown like this: #473   That results in HTML like this:

```
<a class="issue-link js-issue-link" data-error-text="Failed to load title" data-id="616627634" data-permission-text="Title is private" data-hovercard-type="issue" data-hovercard-url="/thephpleague/commonmark/issues/473/hovercard" href="https://github.com/thephpleague/commonmark/issues/473" aria-label="#473, Feature Request: Allow InlineMentionParser to be a little more dynamic">#473</a>
```

Populating those additional attributes requires access to the resulting `Link` node which wouldn't be possible if we used the approach above.  So I do think that having the `MentionLinkGeneratorInterface` return a `Link` is the most flexible option.

So to me, the question is really whether we should modify `CallbackLinkGenerator` to force the callable to accept some kind of "`Mention` DTO" to pass that data back and forth.  And if so, should the callable instead be something that implements a new `MentionModifierInterface` or something like that to further ensure enforce types?

We certainly could, but at that point, I feel like our abstraction is a bit out of control - we're now using ~4-5 classes/interfaces to do something that can easily be done with only 10-20 lines in a single custom class that implements `InlineParserInterface`.

> I'm just worried that if we stick with arbitrary parameters (passed by reference), it will likely eventually bite us in the ass if it ever needs to be expanded upon.

I completely understand where you're coming from here.  Personally, I can't think of a situation where this would happen right now.  But if we ever did need to expand upon it, I'd make sure it would be done in a backward-compatible manner - either by adding additional parameters or deprecating the `CallbackLinkGenerator` in favor of something more suited to whatever the new needs would be.

In other words, I think this implementation is a decent compromise between flexibility and ease of use, and we still have the option to tweak things later in a way that maintains BC without introducing any nasty surprises.

Issue #474, opened 2020-05-14
Feature Request: Merge Attributes and Footnotes Extensions


markcarver:
Given that this project has now consolidated a ton of these rando extensions into itself, I think the last two remaining popular extensions should follow suit:

- [ ] https://github.com/webuni/commonmark-attributes-extension
- [x] https://github.com/rezozero/commonmark-ext-footnotes

x-ref:
https://github.com/rezozero/commonmark-ext-footnotes/issues/7

colinodell:
I agree with footnotes, not sure about the attributes extension though.  While I definitely see the value in having that available here, I'm also very cautious about adding anything that's not rooted within the official CommonMark or GFM specs or implementations.

markcarver:
Well, going by that rationale, then the footnotes extension isn't in either... either.

FWIW, [attributes](https://packagist.org/packages/webuni/commonmark-attributes-extension) has over 88k installs, compared to [footnotes](https://packagist.org/packages/rezozero/commonmark-ext-footnotes) which has just over 1.3k installs. It's an extension I use all the time if only to do simple things like assigning IDs to headers.

While I understand the desire to not conflate spec implementations, these are just extensions on top of the spec (which is essentially what GFM is as well)

If anything this would only help get it [accepted as part of the official spec](https://talk.commonmark.org/t/consistent-attribute-syntax/272) faster.

markcarver:
Also, #421 isn't either...

colinodell:
True, the specs don't have them, but the GFM _implementation_ does have [footnotes](https://github.com/github/cmark-gfm/blob/master/src/plaintext.c#L194), and their Markdown pipeline does support [emojis](https://github.com/github/cmark-gfm/issues/75#issuecomment-355440087).  For those reasons, I feel like it's reasonable to assume these are as close to a standard as we have right now, and any future official specification of those will likely be based closely on those implementations.

I'm still open to the idea of attributes, but I'd like to better understand some of the various implementations other Markdown flavors are using and where the CommonMark community might be leaning.

colinodell:
Maybe for the attributes, we could provide different "flavors" of them?  That way, if an official variant comes out, we don't have to break BC - we just offer the new official flavor?

markcarver:
> but the GFM _implementation_ does have [footnotes](https://github.com/github/cmark-gfm/blob/master/src/plaintext.c#L194)

Still, not part of the spec.

> Emoji is not implemented as part of our Markdown stack

[...]

PR #484, opened 2020-05-23
Footnote Extension

colinodell:
This extension is based on https://github.com/rezozero/commonmark-ext-footnotes, imported and relicensed with permission from the maintainer: https://github.com/thephpleague/commonmark/issues/474#issuecomment-631425455

In addition to importing the functionality, a number of configuration options were added, as well as some other small tweaks.


src/Extension/Footnote/Parser/AnonymousFootnoteRefParser.php
```diff
@@ -0,0 +1,62 @@
...
+
+        $cursor->restoreState($state);
+
+        return false;
+    }
+
+    private function createReference(string $label): Reference
+    {
+        $refLabel = Reference::normalizeReference($label);
+        $refLabel = \mb_strtolower(\str_replace(' ', '-', $refLabel));
```
markcarver, line 57: FWIW, this is why I was originally saying that creating a more generic `StubGenerator` (that isn't tied to headings) might be useful; for cases like this.

ambroisemaupate: You should cleanup reference not to get weird HTML class and id : 

https://github.com/rezozero/commonmark-ext-footnotes/blob/72afb206b9b804c7d788815e56af2caf37d6bd1c/src/Parser/AnonymousFootnoteRefParser.php#L47

markcarver: That's why I was suggesting using the `SlugGenerator`, but moving it out of the `HeadingPermalink` namespace and into a more generic one since it could be used by multiple things: https://github.com/thephpleague/commonmark/blob/f40f6194960c60893537b86bf64f2f3369b72f01/src/Extension/HeadingPermalink/SlugGenerator/DefaultSlugGenerator.php#L33-L45

Also, it probably wouldn't hurt to have a similar `slug_generator` option for this extension as well.

colinodell: In #481 we changed the interface to accept a `Node` instead of a `string` in order to provide extra context about the heading node:

https://github.com/thephpleague/commonmark/blob/f40f6194960c60893537b86bf64f2f3369b72f01/src/Extension/HeadingPermalink/SlugGenerator/SlugGeneratorInterface.php#L16-L22

This suggestion sounds like we want to lean back towards using the text as the primary source of information for the slug.

What if we moved this out of the `HeadingPermalink` namespace as suggested and changed the method signature to look more like this:

```php
    /**
     * @param string $text    The text to slugify
     * @param mixed  $context Additional context about the text being slugified (optional)
     */
    public function generateSlug(string $text, $context = null): string;
```

In the case of heading permalinks, `$context` would receive the `Heading` node.  In this particular case we only have a string, so `$context` would be `null`.  And in other potential future cases, we might have `$context` be some other object, or perhaps an array of relevant items.

What would you think about something like that @markcarver?  I'm open to other ideas as well.

colinodell: (Personally I don't like having the `mixed` type here, but I feel like we either do that or have separate interfaces for slugifying strings vs nodes vs whatever else we might have in the future)

markcarver: @colinodell IIRC, that was my original suggestion, just passing an extra parameter for context. In that case, it was the Heading... I'm assuming we wouldn't pass anything here? Unless we wanted to pass the inline parser(s) as well, just so one knows what requested the slug generation.

What if it had two methods (for ease of use):

```php
    public function generateSlug(string $text, $context = null): string;
    public function generateSlugFromNode(Node $node): string;
```

The latter invoking the former, and passing the `Node` as `$context`?

colinodell: > IIRC, that was my original suggestion, just passing an extra parameter for context.

Gotcha, I apologize for missing/misunderstanding that :upside_down_face: 

> In that case, it was the Heading... I'm assuming we wouldn't pass anything here?

Exactly.

> Unless we wanted to pass the inline parser(s) as well, just so one knows what requested the slug generation.

I think we'd be okay omitting that if we implement the `slug_generator` option.

Let me merge this as-is and create a separate MR for these slug-related changes :wink:

markcarver:
I've added support for this in the Drupal Markdown module:

https://git.drupalcode.org/project/markdown/commit/0749113

Screen Shot 2020-05-24 at 1 46 55 PM

PR #486, opened 2020-05-24
Standardize text normalization

### LINKED ISSUE #485: Standardize text normalization
(Pulling https://github.com/thephpleague/commonmark/pull/484#discussion_r429586004 into a separate thread)

There are a few places within this library where text is manipulated into a "normalized" version:

 - [Normalizing reference labels](https://github.com/thephpleague/commonmark/blob/09e9079efc0a7b00d0739961ef46fedfee01006b/src/Reference/Reference.php#L75)
 - [Normalizing footnote labels](https://github.com/thephpleague/commonmark/blob/09e9079efc0a7b00d0739961ef46fedfee01006b/src/Extension/Footnote/Parser/AnonymousFootnoteRefParser.php#L56-L58)
 - [Generating slugs for heading permalinks](https://github.com/thephpleague/commonmark/blob/09e9079efc0a7b00d0739961ef46fedfee01006b/src/Extension/HeadingPermalink/SlugGenerator/DefaultSlugGenerator.php#L23)

In some cases we want these to be "pluggable" so that users can customize the output to fit their needs; in others we want to use a [standardized approach](https://spec.commonmark.org/0.29/#matches) to comply with the spec.  But in both cases I think we can use a common interface for normalizing text.

I'd therefore to unify these under a single interface:

```php
<?php

namespace League\CommonMark\Normalizer;

/**
 * Creates a normalized version of the given input text
 */
interface TextNormalizerInterface
{
    /**
     * @param string $text    The text to normalize
     * @param mixed  $context Additional context about the text being normalized (optional)
     */
    public function normalize(string $text, $context = null): string;
}
```

`$context` can be used to provide additional information about the text being normalized. For example, when [normalizing heading content](https://github.com/thephpleague/commonmark/blob/09e9079efc0a7b00d0739961ef46fedfee01006b/src/Extension/HeadingPermalink/HeadingPermalinkProcessor.php#L87-L92) within the `HeadingPermalinkProcessor`, we can pass along the `Heading` node as the `$context` in case somebody wants to check the heading level when generating a slug.  In other cases this may be `null`.

We'll provide two implementations out-of-the-box:

1. A `TextNormalizer` based on `Reference::normalizeReference` which performs the normalization prescribed by the spec
2. A `SlugNormalizer` which works similarly, but converts text to underscore and uses `-` separators instead of spaces

Existing code will be updated to use these and we'll set relevant deprecation notices where needed.

Furthermore, we'll rename the new `heading_permalink/slug_generator` option to `heading_permalink/slug_normalizer`.

The `FootnoteExtension` will also be modified to use this new normalizer approach, though further research will be needed to determine if we should leverage the spec's reference normalization approach and/or allow customization.

colinodell:
Closes #485


src/Normalizer/TextNormalizer.php
```diff
@@ -0,0 +1,51 @@
...
+     * {@inheritdoc}
+     */
+    public function normalize(string $text, $context = null): string
+    {
+        // Collapse internal whitespace to single space and remove
+        // leading/trailing whitespace
+        $text = \preg_replace('/\s+/', ' ', \trim($text));
+
+        if (!\defined('MB_CASE_FOLD')) {
+            // We're not on a version of PHP (7.3+) which has this feature
```
GrahamCampbell, line 43: We should really have the same behaviour on all php versions? Maybe some of the code to do this can be copied over from https://github.com/symfony/polyfill/blob/master/src/Mbstring/Mbstring.php#L272?

GrahamCampbell: Oh, this is literally what you have done! Ha, I should go to sleep. ;)

colinodell: Hehe :upside_down_face: 

The only inconsistency was that we were converting things to upper-case, whereas PHP 7.3+ seems to use lower-case, so I've made that change.  But everything else matches the polyfill.

There are still some edge cases [that even Symfony's polyfill doesn't handle](https://github.com/symfony/polyfill/issues/173) but it should be close enough.  I don't think it's enough to warrant a bump in the minimum PHP version in 2.x (we're going from 7.1 to 7.2 as of now) as this logic here is "good enough" for those users.

src/Reference/Reference.php
```diff
@@ -71,20 +63,13 @@ public function getTitle(): string
...
-        if (!\defined('MB_CASE_FOLD')) {
-            // We're not on a version of PHP (7.3+) which has this feature
-            $string = \str_replace(self::$caseFold[0], self::$caseFold[1], $string);
-
-            return \mb_strtoupper($string, 'UTF-8');
-        }
+        @trigger_error(sprintf('%s::normlizeReference() is deprecated; use %s::normalize() instead', self::class, TextNormalizer::class), E_USER_DEPRECATED);
 
-        return \mb_convert_case($string, \MB_CASE_FOLD, 'UTF-8');
+        return (new TextNormalizer())->normalize($string);
```
markcarver, line 73: Considering that these don't actually need class instantiation (new), wouldn't making the public methods static make more sense here? In fact, that's what the deprecated message even says: `Use TextNormalizer::normalize() instead`.

src/Normalizer/TextNormalizer.php
GrahamCampbell: Yeh. :)

src/Reference/Reference.php
colinodell: > Considering that these don't actually need class instantiation (new), wouldn't making the public methods static make more sense here?

We need to keep non-static methods so they can be used in configuration options, but I don't see why we couldn't have both a static and non-static method to cater to both use cases.  Give me a few minutes and I'll push up this change :)

> In fact, that's what the deprecated message even says: Use TextNormalizer::normalize() instead.

I'm just following the common PHP convention of referencing all class methods using `::`, even the non-static ones.

colinodell: Added via 25382406

markcarver: > We need to keep non-static methods so they can be used in configuration options

I'm at a loss for why instantiating classes into instances is even necessary. What does it bring to the table?

Configurations can easily be adjusted to account for static classes:

```php
'slug_normalizer' => '\\League\\CommonMark\\Normalizer\\SlugNormalizer\\SlugNormalizer(),
```

You can still keep interface support and strict adherence, just using `is_subclass_of` instead of `instanceof`.

FWIW, even if you instantiate a class into an instance, it would still be possible for the class instances to invoke its static methods using either the `->` notation or the `::`, you don't have to create instance method wrappers that invoke the static "varient" i.e.:

```php
$result1 = $instance->staticMethod();
$result2 = $instance::staticMethod();
$this-assertSame($result1, $result2);
```
We do it all the time in Drupal.

colinodell: The more I think about this, the more I'm against using static methods here.  I think I might revert 2538240.

Static methods make sense for internal uses where we won't be swapping out implementations, like within `Reference`.  But when it comes to things like `slug_normalizer`, I think it's important that we require non-static implementations.

For example, somebody may want to swap out our naive `SlugNormalizer` with https://github.com/cocur/slugify.  That library is not used statically and incurs some overhead when instantiating and configuring it, so we should allow users to reuse an already-instantiated object.  Sure, those users could leverage the singleton pattern to allow static calls to this non-static instance, but that's not a practice I want to encourage.

But going back to the static usages, I think we can avoid them entirely by moving the responsibility for normalizing text out of the `Reference` object and into whatever instantiates them (currently these are all parser and listener instances).  This is probably a better refactoring than making everything static or allowing both static and non-static methods.

> FWIW, even if you instantiate a class into an instance, it would still be possible for the class instances to invoke its static methods using either the -> notation or the ::, you don't have to create instance method wrappers that invoke the static "varient"

TIL!  That would definitely work here if we wanted to go static, but again, the more I think about it, the less I feel like static methods are the best choice.

Let me play with both ideas (yours and mine) to see how they look in practice.  Both approaches have their merits, and once I've tinkered with both I'll have a better idea of whether my argument here is valid :wink:

colinodell: Upon further review, it seems that normalizing the `$label` in the `Reference` constructor is unnecessary because the `ReferenceMap` is where that normalization is critical, and it already normalizes the label in each of its methods.  Furthermore, there is no code anywhere in this library that depends on `Reference` containing a normalized label.

Therefore, the only calls to normalizers are happening within other instantiated objects that make repeated calls to the normalizer.  This eliminates the compelling use case to make them statically callable.

I've therefore dropped 2538240 from this PR and removed normalization from `Reference::__construct()`.

markcarver: > For example, somebody may want to swap out our naive `SlugNormalizer` with https://github.com/cocur/slugify. That library is not used statically and incurs some overhead when instantiating and configuring it, so we should allow users to reuse an already-instantiated object. Sure, those users could leverage the singleton pattern to allow static calls to this non-static instance, but that's not a practice I want to encourage.

I think I pretty much rubber-ducked myself while replying to this 😃  Basically, I think I've come to the same conclusion as what you stated above. Regardless of where an external library may be implemented, any external library instance that is setup/configured would be inaccessible from a static method. With that in mind, I think its safe to say that this is done.

PR #493, opened 2020-06-13
Allow installation on PHP 8; switch to Github Actions

colinodell:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,76 @@
...
+              with:
+                  php-version: ${{ matrix.php }}
+                  extensions: mbstring
+                  coverage: pcov
+
+            - name: Get composer cache directory
+              id: composer-cache
+              run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            - name: Cache dependencies
```
GrahamCampbell, line 34: If you switch to composer 2, you can delete caching all together, since it actually makes things slower.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,76 @@
...
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                php: ['7.1', '7.2', '7.3', '7.4']
+                coverage: [true]
+                composer-flags: ['']
+                include:
+                    - php: '8.0'
+                      coverage: false
+                      composer-flags: '--ignore-platform-reqs'
```
GrahamCampbell, line 19: Better to use `--ignore-platform-req=php` (requires composer 2).

.github/workflows/tests.yml
GrahamCampbell: NB https://github.com/laravel/framework/pull/33158

colinodell: Awesome! I'll get that implemented :)

.github/workflows/tests.yml
colinodell: Thank you for the tip!

Issue #442, opened 2020-04-13
Feature Request: YAML Front Matter extension


colinodell:
[YAML front matter](https://jekyllrb.com/docs/front-matter/) is commonly used to provide additional metadata about a Markdown file.  Why not include that functionality here as an optional extension?  We could expose the parsed data through the root `Document` node's `$data` array.

Taking this a step further - if the front matter contains a `configuration` key, merge its contents into the current environment's configuration, so that you can adjust the parser's behavior on-the-fly on a document-by-document basis?  This could be really powerful!

Imagine we implement the Table of Contents extension in #441, and you have a Markdown file like this:

```md
---
title: This Is My Awesome Blog Post
configuration:
    table_of_contents:
        max_level: 2
---

# Some heading with level 1

## Some heading with level 2

### Some heading with level 3
```

The `configuration` values in the YAML front-matter would automagically tell the Table of Contents extension to only include the first two headers in the ToC.

(For security reasons, that `configuration` override behavior would be disabled by default, but we'd expose a simple option to enable it when instantiating your converter/environment and converting inputs that you trust).

colinodell:
Until #447 is resolved, being able to override configuration settings per-document could lead to some nasty side effects, so I'm punting this feature to 2.0.

markcarver:
Oooh. Me likey 😃 

I actually spent quite a bit of time implementing this in Drupal core as a standalone component, the latest patch is here: https://www.drupal.org/project/drupal/issues/3064854#comment-13339277

Despite the bikeshedding and political partisanship against front-end work, it's quite stable and thoroughly tested.

p.s. a lot of work has gone into the markdown module to finally provide the proper APIs and UI for extensions: https://www.drupal.org/project/markdown/issues/2952435#comment-13580075

> For security reasons, that `configuration` override behavior would be disabled by default, but we'd expose a simple option to enable it when instantiating your converter/environment and converting inputs that you trust

Do you mean something like Drupal's schema types? Where we can specify which parameters are allowed and what types they should be? i.e.

```yaml
title
  type: string
configuration:
  type: mapping
  mapping:
    table_of_contents:
      type: mapping
      mapping:
        max_level:
          type: integer
```

If so, that would be fantastic as it should be relatively easy to convert what we're already doing:
https://git.drupalcode.org/project/markdown/-/blob/26fdcbfdcca58961d265de32614962f7fac9cb7a/config/schema/markdown.schema.yml#L45-73

hason:
[...]

PR #497, opened 2020-06-13
[2.0] Front matter extension

colinodell:
Implements the Front Matter extension (#442)

The front matter is parsed before the Markdown parsing occurs and gets stashed on the `Document` object.  Later on, we use a new `DocumentRenderedEvent` listener to pull that parsed YAML out and ensure it gets returned via the `convertToHtml()` output.

## Sample Usage

```php
<?php

use League\CommonMark\Extension\FrontMatter\FrontMatterExtension;
use League\CommonMark\Extension\FrontMatter\RenderedContentWithFrontMatter;

// Obtain a pre-configured Environment with all the CommonMark parsers/renderers ready-to-go

// Add the extension
$environment->addExtension(new FrontMatterExtension());

// Set your configuration if needed
$config = [
    // ...
];

// Instantiate the converter engine and start converting some Markdown!
$converter = new CommonMarkConverter($config, $environment);

$markdown = <<<MD
---
layout: post
title: I Love Markdown
tags:
  - test
  - example
---

# Hello World!
MD;

$result = $converter->convertToHtml($markdown);

// Check if front matter is present:
if ($result instanceof RenderedContentWithFrontMatter) {
    $frontMatter = $result->getFrontMatter();
}

// Output the HTML using any of these:
echo $result;               // implicit string cast
// or:
echo (string) $result;      // explicit string cast
// or:
echo $result->getContent();
```

## :warning:  &nbsp; Breaking Change

The main `convertToHtml()` method of the `CommonMarkConverter` class no longer returns a `string`; instead, it returns an object that provides the HTML and other optional bits of information, like the parsed YAML front matter.

This BC-break can be mitigated by casting the result of `convertToHtml()` to a `string`, thus allowing consumers to safely call that method on any 1.x or 2.x version.

To help ensure future compatibility, this return object implements Symfony's polyfill for PHP 8's `Stringable` interface, thus allowing consumers to type-hint against that in 2.x and future versions.

I tried other implementations to avoid this BC-break but ultimately found them to be too kludgy and not developer-friendly.

## Future Scope

This implementation does not include the config override functionality.  I've got it working locally but we'll need to address https://github.com/thephpleague/commonmark/issues/442#issuecomment-615895227 first.  We'll save that config override stuff for a near-future PR :wink:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,7 @@ jobs:
             - run: vendor/bin/phpcs
 
     phpunit:
-        name: PHPUnit on ${{ matrix.php }}
+        name: PHPUnit on ${{ matrix.php }} ${{ matrix.composer-flags }}
         runs-on: ubuntu-latest
         strategy:
             matrix:
@@ -46,6 +46,9 @@ jobs:
                     - php: '8.0'
                       coverage: false
                       composer-flags: '--ignore-platform-reqs'
+                    - php: '7.2'
+                      coverage: false
+                      composer-flags: '--prefer-lowest'
 
         steps:
             - uses: actions/checkout@v2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ See <https://commonmark.thephpleague.com/2.0/upgrading/> for detailed informatio
    - `InlineParserEngineInterface`
    - `MarkdownParserState`
    - `MarkdownParserStateInterface`
+   - `RenderedContent`
+   - `RenderedContentInterface`
  - Added several new methods:
    - `FencedCode::setInfo()`
    - `Heading::setLevel()`
@@ -40,6 +42,7 @@ See <https://commonmark.thephpleague.com/2.0/upgrading/> for detailed informatio
 
 ### Changed
 
+ - `CommonMarkConverter::convertToHtml()` now returns an instance of `RenderedContentInterface`. This can be cast to a string for backward compatibility with 1.x.
  - Moved and renamed several classes - [see the full list here](https://commonmark.thephpleague.com/2.0/upgrading/#classesnamespaces-renamed)
  - Implemented a new approach to block parsing. This was a massive change, so here are the highlights:
    - Functionality previously found in block parsers and node elements has moved to block parser factories and block parsers, respectively ([more details](https://commonmark.thephpleague.com/2.0/upgrading/#new-block-parsing-approach))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ See <https://commonmark.thephpleague.com/2.0/upgrading/> for detailed informatio
 
 ### Added
 
+ - Added new `FrontMatterExtension` ([see documentation](https://commonmark.thephpleague.com/extensions/front-matter/))
  - Added new `HtmlFilter` and `StringContainerHelper` utility classes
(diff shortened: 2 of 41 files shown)


docs/1.4/extensions/front-matter.md
```diff
@@ -0,0 +1,176 @@
+---
```
colinodell, line 1: This shouldn't be in the 1.4 folder...

It also contains the configuration stuff that isn't implemented in this PR.  Need to remove this file.

src/Extension/FrontMatter/FrontMatterParser.php
```diff
@@ -0,0 +1,63 @@
...
+final class FrontMatterParser
+{
+    /**
+     * @var FrontMatterParserInterface
+     *
+     * @psalm-readonly
+     */
+    private $yamlParser;
+
+    private const REGEX_FRONT_MATTER = '/^---\\n.*\\n---\n/s';
```
markcarver, line 30: From my work on https://www.drupal.org/project/drupal/issues/3064854, the following regex is probably a little more forgiving in how newlines can be interpreted across OS:

```php
/**
 * The separator used to indicate front matter data.
 *
 * @var string
 */
 const SEPARATOR = '---';

/**
 * The regular expression used to extract the YAML front matter content.
 *
 * @var string
 */
const REGEXP = '/\A(' . self::SEPARATOR . '(.*?)?\R' . self::SEPARATOR . ')(\R.*)?\Z/s';
```

src/Extension/FrontMatter/FrontMatterParser.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        // Trim the last 4 characters (ending ---s and newline)
+        $frontMatter = \substr($frontMatter, 0, -4);
+
+        // Parse the resulting YAML data
+        $data = $this->yamlParser->parse($frontMatter);
+
+        // Store the parsed data in the Document
+        $event->getDocument()->data['front_matter'] = $data;
```
markcarver, line 56: It would be really nice if this also captured the line offset of the front matter.

```php
// Determine the real source line by counting all newlines in the first
// match (which includes the front matter separators) and append a new
// line to denote that the content should start after it.
$line = preg_match_all('/\R/', $frontMatter . "\n") + 1;
```

src/Extension/FrontMatter/FrontMatterParser.php
```diff
@@ -0,0 +1,63 @@
...
+
+        // Advance through any remaining newlines which separated the front matter from the Markdown text
+        $cursor->match('/^\n+/');
```
markcarver, line 59: I'd still use `'/^\R+/'` here.

colinodell: Newlines are already normalized to `\n` characters at this point thanks to the `MarkdownInput` class:
 https://github.com/thephpleague/commonmark/blob/0125ec4c442e61954b0c4c964ea50448e4067fde/src/Input/MarkdownInput.php#L83

This normalization occurs just before the event we're listening to is fired, thus guaranteeing that we'll only have Linux newlines for the rest of the parsing: 

https://github.com/thephpleague/commonmark/blob/71e723e23f41a9ceac34941cf87953eb3009ce37/src/Parser/MarkdownParser.php#L109

src/Extension/FrontMatter/FrontMatterParser.php
colinodell: Implemented via ec7c4615

Issue #500, opened 2020-06-13
Documentation improvements


colinodell:
A working list things I'd like to see change in the docs:

 - [x] Make it easier to see all the extensions in one table near the top of the page.  Right now we have two separate tables near the bottom.
 - [ ] Split the 2.0 upgrade docs into two sections: one for "consumers" and "integrators" who only need to know the most important bits, and another for "customizers" who write their own parsers/renderers/etc and need all the fine details.

markcarver:
I can probably work on the first item in #499

colinodell:
Yeah, if you don't mind, that would be great!

PR #499, opened 2020-06-13
[WIP] Documentation updates

markcarver:

PR #222, opened 2016-01-16
Resolve parent class

colinodell:
Implements #209

PR #501, opened 2020-06-14
Add a proper configurable MentionExtension

markcarver:
~~While working on #499, I realized that having `MentionParser` in the `Extension` namespace doesn't make a whole lot of sense code or documentation-wise.~~

~~This essentially moves the `MentionParser` login from the `Extension` namespace to the `Inline\Parser` namespace (where it belongs). It also moves the "Mention" documentation to the customize section and then introduces three new "Autolink" supplemental extensions (in the autolink namespace) that take advantage of the new `MentionParser`.~~

Extensions should be able to be added using `$environment->addExtension()` with minimal effort.

This essentially creates a new configurable `MentionExtension` where mention parsers can be defined via configuration.

This cleans up the documentation a bit more to be inline with point two of #500:

- customizers: (Inline parsers, renderers, etc.)
- consumers/integrators: Extensions (configurable/preconfigured ^)

---

This also allows the callback generator to return something other than a link. Specifically thinking of access based mentions where anonymous users wouldn't be able to see the mentions, but authenticated users could.

---

Changes the parameter passed to the generators to a single `Mention` object per the original discussion in https://github.com/thephpleague/commonmark/pull/482/files/29f74234c65b421072b6d326b5679e1f326071ce#r426203599 and then subsequently after in Slack.


colinodell:
> having `MentionParser` in the `Extension` namespace doesn't make a whole lot of sense code or documentation-wise. Extensions should be able to be added using `$environment->addExtension()` with minimal effort.

You're absolutely right, there should be _some_ kind of out-of-the-box functionality you can leverage by calling `addExtension()`.  The current implementation does not have this.

Mentions are an interesting case because they will almost always need to behave differently depending on which site is using them.

> It also ... introduces three new "Autolink" supplemental extensions (in the autolink namespace) that take advantage of the new MentionParser.

That would certainly solve that problem :)

Might I propose a possible alternative?  What if we instead allowed mention formats to be defined via configuration?  Right now we have two shortcuts for creating these parsers:

```php
public static function createWithStringTemplate(string $symbol, string $mentionRegex, string $urlTemplate): MentionParser
{
    return new self($symbol, $mentionRegex, new StringTemplateLinkGenerator($urlTemplate));
}

public static function createWithCallback(string $symbol, string $mentionRegex, callable $callback): MentionParser
{
    return new self($symbol, $mentionRegex, new CallbackLinkGenerator($callback));
}
```

So what if we allowed for something like this:

```php
$environment->addExtension(new MentionExtension());
$environment->setConfig([
    'mentions' => [
        'twitter' => [
            'symbol' => '@',
            'regex' => '/^[A-Za-z0-9_]{1,15}(?!\w)/',
            'generator' => 'https://twitter.com/%s',
        ],
        // or perhaps...
        'drupal_user' => [
            'symbol' => '@',
            'regex' => '/^[A-Za-z0-9_\-]+(?!\w)/', // or whatever
            'generator' => $someCallable,
        ],
    ]
]);
```

If `generator` is a string we pass that configuration through `MentionParser::createWithStringTemplate()`.  Otherwise if it's a `callable` we use `MentionParser::createWithCallback()`.

That would allow it to behave as a true extension.

> This essentially moves the MentionParser login from the Extension namespace to the Inline\Parser namespace (where it belongs).

I agree in the context of the current 1.x namespacing scheme. Unfortunately, in 2.0 it'll have to go back into the `Extension` namespace because of how we reorganized the namespaces.  In a nutshell, there are only 4 types of elements that the core (aka everything outside of the `Extension` namespace) should know about:

 - Document
 - Paragraph
 - Text
 - Newline

Everything else (like parsing links) is considered an extension.  I think mentions should be treated the same way in 2.0.  For that reason, I'd prefer to avoid this particular change if at all possible.

> This also allows the callback generator to return something other than a link. Specifically thinking of access based mentions where anonymous users wouldn't be able to see the mentions, but authenticated users could.

I love it!  I haven't dug into your proposed implementation too deeply yet, but being able to return any `AbstractInline` or `null` sounds perfect to me.

colinodell:
Three other thoughts:

1. Although we're in a feature freeze I want to ensure we get the MentionParser stuff right.
2. If we go with a configuration-based approach I think we should remove the `createTwitterHandleParser()`, `createGitHubHandleParser()`, and `createGitHubIssueParser()` methods and instead provide configuration examples in the docs.
3. I really want to keep as much as we can in the `Extension` namespace to make the transition to 2.0 smoother.

I'll take a closer look through the code now and let you know if I have any other thoughts there.

docs/1.5/extensions/autolinks.md
```diff
@@ -31,9 +31,35 @@ $converter = new CommonMarkConverter([], $environment);
...
+// Note: this extension and the above cannot be used at the same time (both based on @ mention symbol).
+$environment->addExtension(new AutolinkGitHubHandleExtension());
+
+// Sample Input:  `#473`
+// Sample Output: `<a href="https://github.com/thephpleague/commonmark/issues/473">#473</a>`
+$environment->addExtension(new AutolinkGitHubIssueExtension('thephpleague/commonmark'));
+```
+
+You can implement you own [custom mention parser](/1.5/customization/mention/) for more advanced uses cases, including
+using callbacks for integrating with internal services.
```
colinodell, line 62: I think we should revert this and stick with the convention in 1.5 and 2.0 that "autolinks" and "mentions" are two separate concepts:

 - Autolinks are things that specifically implement the [GFM autolink extension](https://github.github.com/gfm/#autolinks-extension-), which bascially makes URLs and email addresses clickable links
 - Mentions are a shortcut for creating links by specifying a certain symbol prefix plus some identifier (id, username, etc) which is then expanded into a longer link.

I originally conflated the two because naming things is hard, but ultimately they are different concepts where one is backed by the GFM spec.  I'd therefore like to keep this new convention consistent throughout the documentation, namespaces, etc.

docs/1.5/extensions/mention.md
```diff
@@ -1,145 +0,0 @@
----
-layout: default
-title: Mention Parser
-description: The MentionParser makes it easy to parse shortened references like @colinodell and #123 to custom URLs
----
-
-# Mentions
-
-Although not a full extension, the `MentionParser` makes it easy to parse shortened mentions and references like `@colinodell` to a Twitter URL or `#123` to a GitHub issue URL.  You can create your own custom syntax by choosing the starting symbol, a regular expression to match against, and any custom URL generator you wish.  Pre-built mention parsers are also available for use.
```
colinodell, line 9: Let's make this a true extension in 1.5 instead :wink:

src/Extension/CommonMarkCoreExtension.php
```diff
@@ -64,6 +64,7 @@ public function register(ConfigurableEnvironmentInterface $environment)
             ->addInlineRenderer(InlineElement\HtmlInline::class, new InlineRenderer\HtmlInlineRenderer(), 0)
             ->addInlineRenderer(InlineElement\Image::class,      new InlineRenderer\ImageRenderer(),      0)
             ->addInlineRenderer(InlineElement\Link::class,       new InlineRenderer\LinkRenderer(),       0)
+            ->addInlineRenderer(InlineElement\Mention::class,    new InlineRenderer\LinkRenderer(),       0)
```
colinodell, line 67: Mentions are not part of the CommonMark spec so this shouldn't be added here.

However, because `Mention extends Link` and we implemented #222, this line may not be needed so long as we specify that our mention extension has a soft dependency on this extension.

src/Inline/Parser/MentionGenerator/CallbackGenerator.php
```diff
@@ -0,0 +1,48 @@
...
+
+use League\CommonMark\Inline\Element\AbstractInline;
+use League\CommonMark\Inline\Element\Mention;
+
+final class CallbackGenerator implements MentionGeneratorInterface
+{
+    /**
+     * A callback function which returns the URL to use, or null if no link should be generated
+     *
+     * @var callable(Mention): ?(AbstractInline|false)
```
colinodell, line 22:
```suggestion
     * @var callable(Mention): ?AbstractInline
```

src/Inline/Parser/MentionGenerator/CallbackGenerator.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    private $callback;
+
+    public function __construct(callable $callback)
+    {
+        $this->callback = $callback;
+    }
+
+    public function generateMention(Mention $mention): ?AbstractInline
```
colinodell, line 31: I think I'd prefer the original method signature of taking in the parsed symbol and identifier and letting this class decide how to create the `Mention` (or possibly some other `AbstractInline`)

src/Inline/Parser/MentionParser.php
```diff
@@ -86,22 +86,6 @@ public static function createWithStringTemplate(string $symbol, string $mentionR
...
-    public static function createGitHubHandleParser(): MentionParser
-    {
-        // RegEx adapted from https://github.com/shinnn/github-username-regex/blob/master/index.js
-        return self::createWithStringTemplate('@', '/^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}(?!\w)/', 'https://github.com/%s');
-    }
-
-    public static function createGitHubIssueParser(string $project): MentionParser
-    {
-        return self::createWithStringTemplate('#', '/^\d+/', "https://github.com/$project/issues/%d");
+        return new self($symbol, $mentionRegex, new CallbackGenerator($callback));
```
colinodell, line 89: :100:

docs/1.5/extensions/autolinks.md
markcarver: Done. This is now a proper `MentionExtension`

docs/1.5/extensions/mention.md
markcarver: Done. This is now a proper `MentionExtension`

src/Extension/CommonMarkCoreExtension.php
markcarver: Ah, I forgot about #222. Yeah, that works.

src/Inline/Parser/MentionGenerator/CallbackGenerator.php
markcarver: I really don't like the idea of passing variables by reference. I really feel that this should pass a `Mention` node as it will likely be what is eventually render. It's up to the generator to supply the mention with a URL or return a different `AbstractInline`. This also allows data to be added to the mention prior to rendering, i.e. what github does for it's mention tooltips:

```
<a class="issue-link js-issue-link" data-error-text="Failed to load title" data-id="126995301" data-permission-text="Title is private" data-hovercard-type="pull_request" data-hovercard-url="/thephpleague/commonmark/pull/222/hovercard" href="https://github.com/thephpleague/commonmark/pull/222" aria-label="#222, Resolve parent class">#222</a>
```

src/Extension/Mention/MentionExtension.php
```diff
@@ -0,0 +1,108 @@
...
+use League\CommonMark\ConfigurableEnvironmentInterface;
+use League\CommonMark\Extension\ExtensionInterface;
+
+final class MentionExtension implements ExtensionInterface
+{
+    public function register(ConfigurableEnvironmentInterface $environment)
+    {
+        $registeredMentions = $environment->getConfig('registered_mentions', []);
+
+        // Retrieve mentions, excluding already registered mentions.
```
colinodell, line 23: We can probably avoid this de-duplication logic since this `register` method is only ever called once per extension instance.  Yes, somebody could potentially add multiple `MentionExtension`s to their environment, but the only negative result would be a slight performance decrease which I'm okay with.

src/Extension/Mention/MentionExtension.php
```diff
@@ -0,0 +1,108 @@
...
+        $environment->mergeConfig([
+            'mentions' => [
+                'github_handle' => [
+                    'symbol'    => '@',
+                    'regex'     => '/^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}(?!\w)/',
+                    'generator' => 'https://github.com/%s',
+                ],
+            ],
+        ]);
+    }
```
colinodell, line 74: I was thinking we could remove this code entirely from the codebase and instead just show these sample configurations in the docs.  It would be less to maintain and also helps users understand the different possibilities.

src/Extension/Mention/Mention.php
```diff
@@ -0,0 +1,92 @@
...
+use League\CommonMark\Inline\Element\Link;
+use League\CommonMark\Inline\Element\Text;
+
+class Mention extends Link
+{
+    /** @var string */
+    private $symbol;
+
+    /** @var string */
+    private $handle;
```
colinodell, line 26: I know I called them `handles` in my code, but I think a better term might be `identifier` because not everything will be a username/handle but rather some sort of unique identifier (which could be a handle or something else entirely, like a numeric ID).  What are your thoughts?

tests/functional/Extension/Mention/MentionExtensionTest.php
```diff
@@ -0,0 +1,199 @@
...
+
+        $converter = new CommonMarkConverter([], $environment);
+
+        $this->assertEquals($expected, $converter->convertToHtml($input));
+    }
+
+    public function testCreateGithubIssueExtension(): void
+    {
+        $input = <<<'EOT'
+This feature was implemented thanks to #473 by Mark Carver.
```
colinodell, line 143: Thanks indeed! :wink:

src/Extension/Mention/MentionExtension.php
markcarver: Ok.

src/Extension/Mention/MentionExtension.php
markcarver: Ok.

src/Extension/Mention/Mention.php
markcarver: How about `match` instead, since that is what is being returned from the regular expression: what it matches.

src/Extension/Mention/Generator/MentionGeneratorInterface.php
```diff
@@ -0,0 +1,25 @@
...
+use League\CommonMark\Inline\Element\AbstractInline;
+
+interface MentionGeneratorInterface
+{
+    /**
+     * @param Mention $mention
+     *
+     * @return AbstractInline|null
+     */
+    public function generateMention(Mention $mention): ?AbstractInline;
```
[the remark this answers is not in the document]
markcarver: I think you're forgetting why this needed to be a callback in the first place and not some interface to be implemented. It's primarily about future-proofing internal/custom integration implementations.

By simply invoking a callback and passing a single an existing object, it allows the namespace checks to be done during runtime and not bound to any type-hinted parameters/return values or namespace declarations. It also means it (technically) doesn't have to create new elements using internal API namespaces:

```php
<?php
public function createUserMention($mention)
{
    // Immediately return if not passed the supported Mention object.
    // By checking this here, it avoids using hard type-hinted parameters
    // and namespaces which can cause a PHP fatal. It allows for external
    // custom implementations to weather potential internal API changes
    // that may occur over time/newer versions.
    if (!($mention instanceof \League\CommonMark\Extension\Mention\Mention)) {
      return null;
    }
    // Blah, blah.
    $mention->setUrl('...');
    $mention->setLabel('...');
    // No creation of new elements necessary.
}
```

> having this partial "maybe use me maybe don't" object

I get what you're saying here, but the majority of the time it's going to be setting the URL for the mention, probably overriding the label. That's why it's being passed as the primary object: this is the 90% use case of mentions and how it's intended to be used.

If someone decides to take it further and implement something a little more advanced by returning something else (using more of the internal APIs), then that's on them.

The primary goal with this callback should be setting the mention's URL and label.

I'd really prefer that to be a strongly typed object for future-proofing reasons as its much easier to do if/else checks of the existence of namespaces/classes, properties, methods over arbitrary scalar values.

> Implementors are then responsible for adding whatever `Link` or `Text` or whatever child nodes they want and return `true` to indicate success, or return `false` to indicate that it's not a valid mention.

Except that there are really no more "implementers" now that it's all config based now. Unless we want to add another "renderer" class property to the mention config which doesn't really sound like a great idea TBH and I think just makes it infinitely more complex than it really needs to be.

Regardless, if that were to happen it would change the fact that the implementers that are using the advanced callback would likely lose the context of what it is they've returned in the first place (i.e. the user, permissions, whatever else happens during that process).

This is the other reason I extended this from Link/AbstractWebResource/AbstractInline/Node though: it allows the callback to stay within the same context of the services they've used to determine what the mention should be and also give them the ability to set additional data on the mention (or whatever AbstractInline it decides to return):

```php
$mention->data['attributes']['data-tooltip'] = '...';
```

All within the same singular implementation, future-proof implementation.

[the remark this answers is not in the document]
colinodell: > I think you're forgetting why this needed to be a callback in the first place and not some interface to be implemented. It's primarily about future-proofing internal/custom integration implementations.

You're right, I did forget that.  My apologies!

I still have significant reservations about passing in a partially-constructed `Mention` object where you can't safely call its methods without getting a `TypeError`.

Perhaps instead of passing in a `Mention` and returning `?AbstractInline` we should instead pass in some kind of builder object like this:

```php
final class MentionBuilder
{
    /** @var string */
    private $type;

    /** @var string */
    private $symbol;

    /** @var string */
    private $identifier;

    /** @var string */
    private $label;

    /** @var string|null */
    private $url;
    
    // TODO: Set the first four based on what we parsed (via constructor).

    // TODO: Also provide getters for the first three and getters + setters for the rest.
}
```

We would pass that object into the callback function so users can call `getSymbol()` or `setUrl()` or whatever else they want.  They would then return `true` to indicate success or `false` if no mention should be parsed.

Once the callback returns, we can pull the values out of the `MentionBuilder` to construct the node(s) as needed.

This kind of approach would allow the callbacks to be oblivious to namespaces and types if they desire, while also ensuring this library has an object representation with types that we can use and which isn't only partially constructed.

What are your thoughts on using that type of approach?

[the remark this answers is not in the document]
markcarver: > I still have significant reservations about passing in a partially-constructed `Mention` object where you can't safely call its methods without getting a `TypeError`.

Where would a TypeError come into play? A simple `instanceof` in the beginning of the callback or any other sort of checks like `property_exists` or `method_exists` on the passed object is sufficient.

This approach provides implementers a way to futureproof their code by allowing BC with any previous objects or limitations the previous object may have. That's why there isn't a type-hint on the `$mention` constructor argument.

It intentionally allows anything to be passed as that first parameter and lets the callback handle compatibility inside itself in a non-fatal manner.

> We would pass that object [`MentionBuilder`] into the callback function so users can call `getSymbol()` or `setUrl()` or whatever else they want. They would then return `true` to indicate success or `false` if no mention should be parsed.
> 
> Once the callback returns, we can pull the values out of the `MentionBuilder` to construct the node(s) as needed.

Ok. So it seems you're on board with passing a single object. Why does it have to be a special "one use only" object though?

Why not just create the object that is intended to be used in the first place? Pass this object off to the callback to provide that missing piece the callback was implemented for in the first place (setting the URL) and then just use that as the child object that's appended.

There are also two other reasons why this will not work:

1. Returning a simple boolean means we'll also lose the ability to return a custom `AbstractInline` objects.
2. No attributes (via data) could be added unless we also add this capability to this `MentionBuilder` object. At which point, we're essentially re-creating all the classes `Mention` already inherits. Why are we reinventing the wheel here by recreating existing code for one-time use?

> This kind of approach would allow the callbacks to be oblivious to namespaces and types if they desire

Check. Which is what the current PR's implementation using the `Mention` object does.

> while also ensuring this library has an object representation with types that we can use

Check. Which again is what the current PR's implementation does; save for that last bit of that paragraph... 

> and which isn't only partially constructed.

Both this PR's `Mention` object and the proposed `MentionBuilder` are only partial constructions.

The reason is: we don't know what the URL is supposed to be. That is why it is a callback in the first place as its entire point of existence is to provide this dynamic URL.

Originally, this callback handled this by returning the URL as a string and assumed it was going to be a `Link` that would be constructed from that.

Since objects are inherently passed by reference, this allows the callback to instead set the URL on the passed object itself. This allows us to simplify the return value expectations of the callback to a singular strongly typed `AbstractInline` which is nullable.

If a URL isn't set, it is essentially the same as returning `null` and saying "this mention isn't a match". If another new `AbstractInline` based object is returned, that is used instead.

All ^ could be summarized in something like:

"Hey, here's the `Mention` object we've created on the assumption you're going to want to use it for 99% of your use-cases. We just need you to set its URL you're supposed to be dynamically generating here, give us a completely new `AbastractInline` object to use instead, or do nothing if this mention isn't a match after all."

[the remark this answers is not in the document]
colinodell: > Where would a TypeError come into play?

Your `Mention` class doesn't call the parent constructor which sets the `$url`.  Therefore, anyone calling `->getUrl()` in the callback will get a type error because it will attempt to return a `null`:

https://github.com/thephpleague/commonmark/blob/4e4170434782f5addb8d531c3f835e6a3e9957a7/src/Inline/Element/AbstractWebResource.php#L29-L35

This is what I mean by the object being only partially constructed.

(I think our static analysis tools in `master` require the parent constructor to be called, too.)

> This approach provides implementers a way to futureproof their code by allowing BC with any previous objects or limitations the previous object may have. That's why there isn't a type-hint on the $mention constructor argument.

Personally I think the best way to offer compatibility is providing a strongly-typed API and following semver so we can **guarantee** compatibility.  Developers should ideally be writing code like this:

```php
$environment->addExtension(new MentionExtension());
$environment->setConfig([
    'mentions' => [
        'twitter' => [
            'symbol' => '@',
            'regex' => '/^[A-Za-z0-9_]{1,15}(?!\w)/',
            'generator' => new class implements MentionGeneratorInterface {
                // TODO: Typed method goes here
            }
        ],
]);
```

If we did this properly, the interface we implement for 1.5 would remain identical to 2.x.

But I'm also sympathetic to your use case where you want to maximize the chance of compatibility, which is why I'm open to implementing that `CallbackGenerator` which provides that strongly-typed, interfaced class that can wrap the compatibility you want.  I'm happy to be flexible to accommodate that, but at the same time I can't justify introducing other code that would violate the strict typing and everything else that other people would rely on.

That's the main reason why I don't want to introduce a `Mention` class that doesn't follow the same guarantees everything else does.

> Ok. So it seems you're on board with passing a single object. Why does it have to be a special "one use only" object though?

Because the proposed `Mention` object violates the guarantees we provide about type safety.  We therefore need something else that encapsulates that type of information.

> Why not just create the object that is intended to be used in the first place?

That would absolutely be my preference, but to do so within the callback function, which I know you want to avoid.

> Returning a simple boolean means we'll also lose the ability to return a custom AbstractInline objects.

You can't do that without having your custom objects having a `use` statement to the namespace that `AbstractInline` lives in.  If I understand you correctly, you're trying to avoid those types of references, so I don't think this would be an applicable use case for using the simple callback approach - implementing `MentionGeneratorInterface` would be better since you have to reference the namespaces anyway.

> No attributes (via data) could be added unless we also add this capability to this MentionBuilder object. At which point, we're essentially re-creating all the classes Mention already inherits.

That's a fair point, but if we can't use a partially-constructed `Mention` and we can't construct a `new Mention()` from within the callback, then I think it's our only option - otherwise we're back to returning a URL `string`.

----

Ultimately I have no problem providing a callback wrapper and I can see how it would be valuable to simplify the most common operations.  I just can't justify breaking certain promises, expectations, and best practices that others rely on.  I also think that perhaps the scope of this callback function might be too broad - if somebody really wants to return custom node types and the whole works then perhaps they should implement `MentionGeneratorInterface` which lets them do all of that, and we keep the callback approach focused on only the most common use cases.

> "Hey, here's the Mention object we've created on the assumption you're going to want to use it for 99% of your use-cases. We just need you to set its URL you're supposed to be dynamically generating here, give us a completely new AbastractInline object to use instead, or do nothing if this mention isn't a match after all."

I'm essentially proposing this should be:

> "Hey, here's a ~Mention~ MentionBuilder object we've created on the assumption you're going to want to use it for 99% of your use-cases. We just need you to set its URL you're supposed to be dynamically generating here, ~give us a completely new AbastractInline object to use instead,~ or do nothing if this mention isn't a match after all. **If you need more flexibility, we have this nice `MentionGeneratorInterface` you can use instead.**"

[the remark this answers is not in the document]
markcarver: We continued this discussion in Slack and came up with a game plan. The latest commit addresses the `TypeError` issue. The rest we discussed and are moving forward with the `Mention` object and explicitly requiring the return of the same passed mention object that was passed, after the URL has been set.

This also includes some better documentation regarding different types of callback implementation and when to use them (completely anonymous vs. interfaced/type-hinted).

markcarver:
@colinodell LGTM 😄

PR #506, opened 2020-06-20
Adds new “symbol” config option and deprecates “inner_contents”

### LINKED ISSUE #505: Replace "inner_contents" config option with "symbol" in Heading Permalinks
With the recent request and decision made in #504, it brought up the point that the `inner_contents` config option essentially allows the same thing.

It currently defaults to [HeadingPermalinkRenderer::DEFAULT_INNER_CONTENTS](https://github.com/thephpleague/commonmark/blob/bd34d0e3dd9a5f7c4bbaaf931536b22d35fe5d6c/src/Extension/HeadingPermalink/HeadingPermalinkRenderer.php#L26) which is an [Octicon link SVG](https://iconify.design/icon-sets/octicon/link.html) element.

### Problems

- It creeps into front-end territory where decisions regarding visual aspects like this should live.
- It bakes the custom markup into the rendered output. This can have serious consequences in CMSes where content is typically only ever rendered at the time of edit/save. This means the rendered output can be archived or cached indefinitely; causing issues during a site redesign where the markup provided may no longer work or in fact, may cause serious display issues.
- Allows any arbitrary HTML and potential future security vulnerability that may arise due to the planned allowance of runtime configuration overrides via front matter (#497).

### Proposed solution

- Introduce a new `symbol` config option that is limited to a single character and defaults to a paragraph symbol (`¶`).
- Deprecate the `inner_contents` config option and remove the default value.
- Only use `inner_contents` if it was explicitly specified in config.
- Update documentation to reflect this change and give additional CSS examples on how to provide a custom icon similar to https://commonmark.thephpleague.com/1.5/extensions/external-links/#adding-icons

markcarver:
Closes #505


docs/1.5/extensions/heading-permalinks.md
```diff
@@ -59,16 +59,41 @@ The value of this nested configuration option should be a `string` that you want
...
 This controls whether the anchor is added to the beginning of the `<h1>`, `<h2>` etc. tag or to the end.  Can be set to either `'before'` or `'after'`.
 
+### `symbol`
+
+This option sets the symbol used to display the permalink on the document. This defaults to `\League\CommonMark\Extension\HeadingPermalink\HeadingPermalinkRenderer::DEFAULT_SYMBOL = '¶'`.
+
+If you want to use a custom icon, then set this to an empty string `''` and check out the [Adding Icons](#adding-icons) sections below.
+
+> Note: the symbol should only be is a single character value; additional characters will be stripped (does not affect multibyte characters).
+
```
colinodell, line 96:
```suggestion
```

docs/1.5/extensions/external-links.md
```diff
@@ -124,22 +124,40 @@ class MyCustomLinkRenderer implements InlineRendererInterface
...
-// Font Awesome example:
+```php
+$config = [
+    'external_link' => [
+        'html_class' => 'external',
+    ],
+];
+```
+
+```scss
```
colinodell, line 137: Not everyone uses SCSS (though I personally think they should!)  so should we change this example to work without needing to use `@extend`?  I'm open to your thoughts here.

docs/1.5/extensions/external-links.md
```diff
@@ -124,22 +124,40 @@ class MyCustomLinkRenderer implements InlineRendererInterface
...
 a.external::after {
-  @extend .glyphicon;
-  content: "\e164";
-  margin-left: .5em;
-  margin-right: .25em;
+  display: inline-block;
+  content: "";
+  // Tip: use an SVG URL encoder (https://yoksel.github.io/url-encoder).
+  // https://icons.getbootstrap.com/icons/link-45deg/
+  background-image: url("data:image/svg+xml,%3Csvg class='bi bi-link-45deg' viewBox='0 0 16 16' fill='currentColor' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M4.715 6.542L3.343 7.914a3 3 0 104.243 4.243l1.828-1.829A3 3 0 008.586 5.5L8 6.086a1.001 1.001 0 00-.154.199 2 2 0 01.861 3.337L6.88 11.45a2 2 0 11-2.83-2.83l.793-.792a4.018 4.018 0 01-.128-1.287z'/%3E%3Cpath d='M5.712 6.96l.167-.167a1.99 1.99 0 01.896-.518 1.99 1.99 0 01.518-.896l.167-.167A3.004 3.004 0 006 5.499c-.22.46-.316.963-.288 1.46z'/%3E%3Cpath d='M6.586 4.672A3 3 0 007.414 9.5l.775-.776a2 2 0 01-.896-3.346L9.12 3.55a2 2 0 012.83 2.83l-.793.792c.112.42.155.855.128 1.287l1.372-1.372a3 3 0 00-4.243-4.243L6.586 4.672z'/%3E%3Cpath d='M10 9.5a2.99 2.99 0 00.288-1.46l-.167.167a1.99 1.99 0 01-.896.518 1.99 1.99 0 01-.518.896l-.167.167A3.004 3.004 0 0010 9.501z'/%3E%3C/svg%3E");
```
colinodell, line 159: Could we use the Octicon link SVG here, and put this example above the others?  Some users may want to stick with the same icon they're using from 1.4 and so having that specific example will help them migrate to 2.x a little easier.

And/or we could add something to the upgrade guide page  showing exactly how to keep the Octicon link in 1.5+.

docs/1.5/extensions/heading-permalinks.md
```diff
@@ -59,16 +59,41 @@ The value of this nested configuration option should be a `string` that you want
...
+$config = [
+    'heading_permalink' => [
+        'inner_contents' => HeadingPermalinkRenderer::DEFAULT_INNER_CONTENTS,
+    ],
+];
+```
 
-By default, we provide an embedded [Octicon link SVG](https://octicons.github.com/icon/link/), but you can replace this with any custom HTML you wish.
+Whenever this config option is provided a value, a deprecation warning will be triggered and the `symbol` config
+option below will be ignored completely.
```
colinodell, line 83: Assuming we add instructions to the upgrading guide, I think we could simply this to:

```suggestion
This controls the HTML you want to appear inside of the generated `<a>` tag.  Usually this would be something you'd style as some kind of link icon, but you can replace this with any custom HTML you wish.

This option was deprecated in 1.5.0 and will be removed in 2.0.0.  Use the `symbol` option instead.  See the [Upgrade Guide](/1.5/upgrading/) for more information.
```

We'll need to provide information in that upgrade guide too, and based on my other comment I think it should include information on how to keep the Octicon link, either as a CSS rule (recommended) or an embedded SVG (the way it is now, not recommended but still supported)

Also, note that we _might_ end up having a 1.4.4 release if there are any security issues, which is why I removed the reference to 1.4.3. Hopefully not, though! :crossed_fingers:

src/Extension/HeadingPermalink/HeadingPermalinkRenderer.php
```diff
@@ -55,8 +58,18 @@ public function render(AbstractInline $inline, ElementRendererInterface $htmlRen
...
+        if ($innerContents !== null) {
+            @trigger_error(sprintf('The %s config option is deprecated; use %s instead', 'inner_contents', 'symbol'), E_USER_DEPRECATED);
+
+            return new HtmlElement('a', $attrs, $innerContents, false);
+        }
+
+        $symbol = $this->config->get('heading_permalink/symbol', self::DEFAULT_SYMBOL);
+        if (mb_strlen($symbol) > 1) {
+            $symbol = mb_substr($symbol, 0, 1);
+        }
```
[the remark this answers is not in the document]
markcarver: Then there's no point introducing this new config option...

docs/1.5/extensions/external-links.md
markcarver: Let's just get rid of all the SCSS and fa/bootstrap examples.

docs/1.5/extensions/external-links.md
markcarver: Let's just provide this one example using the Octicon link SVG in plain CSS. The main point of this is to show how to target and style via pseudo-selector. If someone chooses to use a different library, that's up to them.

docs/1.5/extensions/heading-permalinks.md
markcarver: Done

src/Extension/HeadingPermalink/HeadingPermalinkRenderer.php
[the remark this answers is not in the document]
colinodell: I think it helps to clarify the purpose of that option and set the expectation that it should probably be a single character.  That being said, the only thing stopping somebody from adding multiple characters is this "arbitrary" restriction, so I think we can safely remove it.

docs/1.4/upgrading.md
```diff
@@ -72,3 +72,52 @@ This class has several unused methods, or methods with an existing alternative:
 The `Converter` class has been deprecated - switch to using `CommonMarkConverter` instead.
 
 The `ConverterInterface` has been deprecated - switch to using `MarkdownConverterInterface` instead.
+
+### `heading_permalink/inner_contents` configuration option
```
colinodell, line 76: Let's move this into `docs/1.5/upgrading.md`.

src/Extension/HeadingPermalink/HeadingPermalinkRenderer.php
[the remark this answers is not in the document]
colinodell: Based on our discussion in Slack, let's run the value through something like `Xml::escape()` or `\htmlspecialchars()` instead of limiting length.

docs/1.5/extensions/heading-permalinks.md
colinodell:
```suggestion
> Note: Any HTML characters provided here will be escaped for security reasons.

```

docs/1.4/upgrading.md
markcarver: Whoops, had both folders open. Done.

PR #515, opened 2020-07-07
[2.0] Nice actions output

GrahamCampbell:
These changes allow the tools to annotate the code with their errors. Note that phpstan already does this as of the most recent patch release 6 days ago, or whenever it was. The other tools need a little bit of help to detect GitHub actions.

The quiet and no-progress flags are not strictly necessary, but they prevent garbled output where progress bars are not rendered correctly (just like you already had).


GrahamCampbell:
Oh, it looks like something's up with PHP 8.0. I'll take a look for a separate PR. :)

PR #516, opened 2020-07-07
[1.5] Nice actions output

GrahamCampbell:
Same as https://github.com/thephpleague/commonmark/pull/515, but for the smaller number of tools that run on 1.5.


GrahamCampbell:
Failing tests address here: https://github.com/thephpleague/commonmark/pull/517.

PR #517, opened 2020-07-07
[1.5] Check class_exists before get_parent_class

GrahamCampbell:
Should fix the tests on PHP 8.0 again. :)

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -410,7 +410,7 @@ final class Environment implements ConfigurableEnvironmentInterface
             return $list[$class];
         }
 
-        while ($parent = \get_parent_class($parent ?? $class)) {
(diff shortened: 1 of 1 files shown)


src/Environment.php
```diff
@@ -410,7 +410,7 @@ private function getRenderersByClass(array &$list, string $class, string $type):
             return $list[$class];
         }
 
-        while ($parent = \get_parent_class($parent ?? $class)) {
+        while (\class_exists($parent = $parent ?? $class) && $parent = \get_parent_class($parent)) {
```
GrahamCampbell, line 413: Kinda funky assignments, but this is the most compact and efficient way to do it I think?

colinodell: Works for me!

Issue #492, opened 2020-06-03
Autolink extension breaks some URLs too early


inguin:
**Version(s) affected**: 1.5

**Description**  
The autolink extension breaks certain URLs too early. We have a Wiki system that uses URLs like `http://wiki/Puncutation_in_links:_why_its_bad_(and_should_be_avoided)`. The markdown parser recognizes the part between the colon and the opening parenthesis as emphasized text. The autolink extension only sees the node with the URL up to the colon.

**How to reproduce**  
```php
$environment = \League\CommonMark\Environment::createCommonMarkEnvironment();
$environment->addExtension(new \League\CommonMark\Extension\Autolink\AutolinkExtension());
$converter = new \League\CommonMark\CommonMarkConverter([], $environment);
echo $converter->convertToHtml('http://wiki/Punctuation_in_links:_why_its_bad_(and_should_be_avoided)');
```

**Possible Solution**  
Not sure if there's an easy fix. I assume the autolink processor could somehow merge subsequent nodes as long as the regex consumes everything.

colinodell:
Yeah, the challenge is that auto-link parsing happens at the very end of the parsing step, where the AST looks something like this before the autolink extension gets it:

```xml
<document>
  <paragraph>
    <text>http://wiki/Puncutation_in_links:</text>
    <emph>
      <text>why_its_bad</text>
    </emph>
    <text>(and_should_be_avoided)</text>
  </paragraph>
</document>
```

The emphasized text could have been created with underscores or asterisks, but we no longer have that information at this point.  We'd probably have to check for links much earlier during the parsing step which would be a significant change, and likely not something we could do within the 1.x branch, but we could investigate a possible solution in 2.0.


----

For what it's worth, we're not the only ones with this problem: https://babelmark.github.io/?text=http%3A%2F%2Fwiki%2FPuncutation_in_links%3A_why_its_bad_(and_should_be_avoided)

markcarver:
It seems to be working with native GFM (which the extension is based on).

Maybe a temporary solution in 1.x could be to create a separate parser that runs much earlier to extract links and replace them with temporary placeholders to be replaced at a later time?

colinodell:
We could add an inline parser that looks for `w` or `h` (since URL autolinks must start with `www` or `http`), but that would have some major performance impacts since the parser would stop at every single `w` and `h` character it sees (most of which would not be autolinks).

Some other potential ways of addressing this:

## Revise inline parsing in 2.0

[...]

PR #518, opened 2020-07-07
[1.5] Fixed GFM autolinking spec compliance

GrahamCampbell:
So, a funny thing has happened where both commonmark.js and GFM forgot to tag releases, despite commonmark.js saying there was a release (0.19.2) and GFM claiming there is a release (0.29.0.gfm.1) too and quoting it in a security vulnerability report from 7 days ago! With all this in mind, I think we should treat their current states as the intended releases, so I have... GFM seem to have changed their autolinking rules to say that all trailing parens should be treated as not part of the link. I have updated the autolink extension as appropriate (indeed, the old implementation failed their tests before):

```
Unexpected result:
=== markdown ===============
www.google.com/search?q=Markup+(business)

www.google.com/search?q=Markup+(business)))

(www.google.com/search?q=Markup+(business))

(www.google.com/search?q=Markup+(business)

=== expected ===============
<p><a␣href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a></p>
<p><a␣href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a>))</p>
<p>(<a␣href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a>)</p>
<p>(<a␣href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a></p>

=== got ====================
<p><a␣href=""></a></p>
<p><a␣href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a>))</p>
<p>(<a␣href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a>)</p>
<p>(<a␣href=""></a></p>

Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'<p><a href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a></p>\n
+'<p><a href=""></a></p>\n
 <p><a href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a>))</p>\n
 <p>(<a href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a>)</p>\n
-<p>(<a href="http://www.google.com/search?q=Markup+(business)">www.google.com/search?q=Markup+(business)</a></p>\n
+<p>(<a href=""></a></p>\n
 '
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
 
 ## [Unreleased][unreleased]
 
+### Changed
+
+ - Bumped CommonMark spec compliance to 0.28.2
+
+### Fixed
+
+ - Fixed `textarea` elements not being treated as a type 1 HTML block (like `script`, `style`, or `pre`)
+
 ## [1.5.4] - 2020-08-17
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
 ### Fixed
 
  - Fixed `textarea` elements not being treated as a type 1 HTML block (like `script`, `style`, or `pre`)
+ - Fixed autolink processor not handling other unmatched trailing parentheses
 
 ## [1.5.4] - 2020-08-17
 
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "require-dev": {
         "ext-json": "*",
         "cebe/markdown": "~1.0",
-        "commonmark/commonmark.js": "0.29.1",
+        "commonmark/commonmark.js": "0.29.2",
         "erusev/parsedown": "~1.0",
         "github/gfm": "0.29.0",
         "michelf/php-markdown": "~1.4",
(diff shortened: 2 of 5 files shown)


colinodell:
I'd like to wait and see the response to https://github.com/commonmark/commonmark.js/issues/190 to determine if 0.29.2 is actually considered a final version of the spec that we should support now, or if we should wait on this.

And thanks for letting me know about [the vulnerability report](https://github.com/github/cmark-gfm/security/advisories/GHSA-7gc6-9qr5-hc85) - I'm only following the releases of that repo and not all activity, which is probably why I didn't see a notification.  It seems to be specific to GFM's table implementation (which ours is not based on) and is probably mitigated by the `max_nesting_level` setting in this project, though it's worth digging into further to confirm.

markcarver:
This is likely is completely or semi-related to #492

colinodell:
CommonMark spec 0.28.2 has been released but no corresponding GFM spec release exists just yet.  But the autolinking change is technically compatible with what the current GFM spec calls for, so I'm okay committing that change (along with our own tests) without bumping the GFM spec to an unreleased version.

I've made the necessary changes and force-pushed to your branch.

GrahamCampbell:
🍻

PR #520, opened 2020-07-18
Allow heading permalink levels to be configurable (#519)

### LINKED ISSUE #519: [Heading Permallink]: Define levels
**Description**  

At present there doesn't seem to be a way to configure the heading permalink extension to ignore some headers - compare to the table of contents extension.

Could be an array of header levels to include mapped to the markdown or html equivalents.

I have a site where the first line of HTML is always at the very top and is an H1 heading and I'd rather it not have this added. I use a workaround at present.

**Example**  

```php
// default
["heading_permalink" => [
  "levels" => [1, 2, 3, 4, 5, 6]
]

// skip h1
["heading_permalink" => [
  "levels" => [2, 3, 4, 5, 6]
]
```

Seems like it would be the most versatile approach.

colinodell:
Resolves #519


src/Extension/HeadingPermalink/HeadingPermalinkProcessor.php
```diff
@@ -58,11 +58,13 @@ public function __invoke(DocumentParsedEvent $e): void
     {
         $this->useNormalizerFromConfigurationIfProvided();
 
+        $levels = $this->config->get('heading_permalink/levels', [1, 2, 3, 4, 5, 6]);
```
GrahamCampbell, line 61: `/** @var int[] */`

markcarver: The ToC extension already has simlar configuration:

```
        'min_heading_level' => 1,
        'max_heading_level' => 6,
```

Can we use the same config min/max here to avoid introducing confusion (trying to remember which is which)?

Or, introduce this new array of `levels` to the ToC extension and deprecate the min/max approach?

colinodell: Good catch.  This has been changed via 2aa08dc3.

PR #521, opened 2020-07-18
Fix tests on PHP 8

colinodell:
"match" is now a semi-reserved keyword in PHP 8, so we need the patch from https://github.com/sebastianbergmann/phpunit/pull/4374 in order for our tests to run.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,8 @@
         "scrutinizer/ocular": "^1.5",
         "symfony/finder": "^4.2"
     },
+    "minimum-stability": "dev",
+    "prefer-stable": true,
     "conflict": {
         "scrutinizer/ocular": "1.7.*"
     },
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "michelf/php-markdown": "~1.4",
         "mikehaertl/php-shellcommand": "^1.4",
         "phpstan/phpstan": "^0.12",
-        "phpunit/phpunit": "^7.5",
+        "phpunit/phpunit": "^7.5 || ^8.5 || ^9.2",
         "scrutinizer/ocular": "^1.5",
         "symfony/finder": "^4.2"
     },
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,6 +30,10 @@ jobs:
 
             - run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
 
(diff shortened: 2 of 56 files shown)


composer.json
```diff
@@ -32,10 +32,12 @@
         "michelf/php-markdown": "~1.4",
         "mikehaertl/php-shellcommand": "^1.4",
         "phpstan/phpstan": "^0.12",
-        "phpunit/phpunit": "^7.5",
         "scrutinizer/ocular": "^1.5",
-        "symfony/finder": "^4.2"
+        "symfony/finder": "^4.2",
+        "symfony/phpunit-bridge": "^5.1"
```
GrahamCampbell, line 37: There is currently a bug effecting phpunit bridge which means it doesn't work with composer 2 and phpunit 9.3.

GrahamCampbell: https://github.com/symfony/symfony/issues/37601

colinodell: Ah, thanks for that!  I've been trying to follow your work over the past day or two on this but didn't come across that one yet :wink:

GrahamCampbell: Yeh, I've mostly not had to deal with the phpunit-bridge package. I came across this bug because Composer itself uses it to run its tests, and ran into the same problem as you: not using PHPUnit 9 yet, but getting match syntax error on PHP 8.

colinodell: phpunit-bridge will help us easily test on PHP 7.1 - 8.0, as those will require different versions of PHPUnit, some of which need `: void()` return types whereas others don't.

Once https://github.com/symfony/symfony/pull/37607 is merged and released we should be able to proceed with this MR here.

GrahamCampbell: > phpunit-bridge will help us easily test on PHP 7.1 - 8.0, as those will require different versions of PHPUnit, some of which need : void() return types whereas others don't.

is that actually needed? You can replace setup and teardown methods with your own methods with annotations.

GrahamCampbell: Also, adding void where the interface didn't want it is permitted? https://3v4l.org/9me7a

colinodell: This is true, but I'm a lazy developer and would rather avoid extra work and merge conflicts if I can :stuck_out_tongue:

GrahamCampbell: Fair enough. Hopefully someone from the Symfony core team will review the PR on Monday.

colinodell: Meh, I have nothing else to do today.  I'll go native PHPUnit and try out the bridge another time.

PR #522, opened 2020-07-18
Allow custom footnote backref link text

bakerkretzmar:
This PR adds a `footnote/backref_text` configuration option to allow users to specify the content of the backref link in footnotes.

Currently the content of this link is hard-coded as `&#8617;`, which should render `↩` (Unicode "Leftwards Arrow with Hook"), but this sometimes displays as the 'right arrow curving left' emoji, ↩️, instead. I'm not sure if this is a browser thing or a macOS thing (I tried latest Firefox and Chrome but don't have access to a Windows machine). This PR would allow setting this link text to any string, like `Go back.`, `↑`, etc.


src/Extension/Footnote/Renderer/FootnoteBackrefRenderer.php
```diff
@@ -43,7 +43,7 @@ public function render(Node $node, ChildNodeRendererInterface $childRenderer)
         $attrs['href']  = \mb_strtolower($node->getReference()->getDestination());
         $attrs['role']  = 'doc-backlink';
 
-        return '&nbsp;' . new HtmlElement('a', $attrs, '&#8617;', true);
+        return '&nbsp;' . new HtmlElement('a', $attrs, $this->config->get('footnote/backref_text', '&#8617;'), true);
```
markcarver, line 46: Thanks, @bakerkretzmar! This definitely should be a configuration option; probably just got overlooked when porting it from the original project (which had none).

One tiny note:

As as done in #506 (https://github.com/thephpleague/commonmark/pull/506/files#diff-5b5681fa1e8206c321de09d716ed6533), this needs to be escaped using `htmlspecialchars`. It should also probably be renamed to `backref_symbol`. If someone needs more complex styling, they could remove this symbol entirely and style it by targeting it via CSS using `backref_class`.

bakerkretzmar: Thanks, updated!

bakerkretzmar:
@markcarver thanks for the feedback! Didn't realize the docs are also in this repo, but I've updated them based on your Heading Permalink PR.

I used the actual unicode character `↩` instead of `&#8617;`, because `htmlspecialchars` breaks that and I see you've done it elsewhere with `¶`.

markcarver:
> Should I update this PR to target the `1.5` branch instead of `latest`?

No, 1.5 is only for non-bc breaking bug fixes and security releases at this point. New features will be added in 2.0.

docs/1.5/extensions/footnotes.md
```diff
@@ -94,6 +95,14 @@ This extension can be configured by providing a `footnote` array with several ne
 
 This `string` option defines which HTML class should be assigned to rendered footnote backreference elements.
 
+### `backref_symbol`
+
+This `string` option sets the symbol used as the contents of the footnote backreference link. It defaults to `\League\CommonMark\Extension\Footnote\Renderer\FootnoteBackrefRenderer::DEFAULT_SYMBOL = '↩'`.
+
+If you want to use a custom icon, set this to an empty string `''` and use CSS to override the elements' contents. See the [Heading Permalink Extension docs](/1.5/extensions/heading-permalinks#adding-icons) for an example of how to do this.
```
markcarver, line 102: Minor nit/personal preference: generally speaking documentation should give examples for each feature/implementation, not reference something else saying it's "similar". While it's true that most could probably extrapolate the inference here, this can be wildly confusing for new developers. Let's just c&p the same example 
 down below and reword it to work for footnotes explicitly.

tests/unit/Extension/Footnote/Renderer/FootnoteBackrefRendererTest.php
```diff
@@ -64,4 +64,18 @@ public function testClassConfiguration(): void
...
+    {
+        $renderer = new FootnoteBackrefRenderer();
+        $renderer->setConfiguration(new Configuration(['footnote' => ['backref_symbol' => 'Go back.']]));
+
+        $fakeReference   = new Reference('label', 'dest', 'title');
+        $footnoteBackref = new FootnoteBackref($fakeReference);
+
+        $output = $renderer->render($footnoteBackref, new FakeChildNodeRenderer());
+
+        $this->assertStringContainsString('Go back.', $output);
```
markcarver, line 78: Minor nit/personal preference: generally, I try to avoid assertions like this as it doesn't guarantee the expected position of where a new configurable element should be injected. In other tests, this kind of assertion may make more sense, but I think this one should assert the entire HTML output to ensure the symbol isn't accidentally and arbitrarily moved.

markcarver: P.S. It should also likely test for special HTML characters to ensure they're escaped properly.

docs/1.5/extensions/footnotes.md
markcarver: Also, these doc changes should be in the 2.0 directory.

bakerkretzmar: Sounds good! Copy+paste was what I was trying to avoid, but I agree it's clearer to just be explicit.

tests/unit/Extension/Footnote/Renderer/FootnoteBackrefRendererTest.php
bakerkretzmar: @markcarver thanks, I'll add a test for special HTML characters. Can you show me an example of what you mean for the first point? I didn't see anything like that for the Heading Permalink extension, and there are a bunch of snapshot and spec tests that do assert the entire output, so here I was mostly just testing that the config option works.

bakerkretzmar:
> No, 1.5 is only for non-bc breaking bug fixes and security releases at this point. New features will be added in 2.0.

Sounds good, should I also target the 2.0 docs then?

markcarver:
> Sounds good, should I also target the 2.0 docs then?

Yes

tests/unit/Extension/Footnote/Renderer/FootnoteBackrefRendererTest.php
markcarver: @bakerkretzmar, sure: https://github.com/thephpleague/commonmark/blob/3b22eb3be0096a20f6edd8ab6a1516d437a9839e/tests/functional/Extension/HeadingPermalink/HeadingPermalinkExtensionTest.php#L45-L74

bakerkretzmar:
@markcarver I took a stab at a similar test to the example you linked, let me know if that's what you wanted!

Along the way I think I discovered a bug... customizing`ref_id_prefix` and `footnote_id_prefix` doesn't appear to work, because references to them are hard-coded as `fnref:` and `fn:` in `AnonymousFootnoteListener.php`, `GatherFootnotesListener.php`, and `FootnoteRefParser.php`:

https://github.com/thephpleague/commonmark/blob/d9fdf5f5ea370c988375e54c85a43b2668a048e6/src/Extension/Footnote/Event/AnonymousFootnotesListener.php#L46

https://github.com/thephpleague/commonmark/blob/d9fdf5f5ea370c988375e54c85a43b2668a048e6/src/Extension/Footnote/Event/GatherFootnotesListener.php#L52

I'll submit this as a new issue and PR, and then the test can be updated or another can be added to make sure customizing those options works too.

bakerkretzmar:
@markcarver done!

colinodell:
Thanks for your contribution!  I've rebased, squashed, and merged this into `latest` for the upcoming `2.0` release :sunglasses: :tada:

bakerkretzmar:
🙌 thanks!

PR #532, opened 2020-08-14
[2.0] Front matter extension

colinodell:
_(Creating a new PR as #497 was accidentally closed)_

TODO:

- [x] Rebase this onto `latest`; resolve merge conflicts
- [x] [Allow for ad-hoc parsing of front matter without rendering](https://github.com/thephpleague/commonmark/issues/442#issuecomment-674285776)

------

Implements the Front Matter extension (#442)

The front matter is parsed before the Markdown parsing occurs and gets stashed on the `Document` object.  Later on, we use a new `DocumentRenderedEvent` listener to pull that parsed YAML out and ensure it gets returned via the `convertToHtml()` output.

## Sample Usage

```php
<?php

use League\CommonMark\Extension\FrontMatter\FrontMatterExtension;
use League\CommonMark\Extension\FrontMatter\RenderedContentWithFrontMatter;

// Obtain a pre-configured Environment with all the CommonMark parsers/renderers ready-to-go

// Add the extension
$environment->addExtension(new FrontMatterExtension());

// Set your configuration if needed
$config = [
    // ...
];

// Instantiate the converter engine and start converting some Markdown!
$converter = new CommonMarkConverter($config, $environment);

$markdown = <<<MD
---
layout: post
title: I Love Markdown
tags:
  - test
  - example
---

# Hello World!
MD;

$result = $converter->convertToHtml($markdown);

// Check if front matter is present:
if ($result instanceof RenderedContentWithFrontMatter) {
    $frontMatter = $result->getFrontMatter();
}

// Output the HTML using any of these:
echo $result;               // implicit string cast
// or:
echo (string) $result;      // explicit string cast
// or:
echo $result->getContent();
```

## :warning:  &nbsp; Breaking Change

The main `convertToHtml()` method of the `CommonMarkConverter` class no longer returns a `string`; instead, it returns an object that provides the HTML and other optional bits of information, like the parsed YAML front matter.

This BC-break can be mitigated by casting the result of `convertToHtml()` to a `string`, thus allowing consumers to safely call that method on any 1.x or 2.x version.

To help ensure future compatibility, this return object implements Symfony's polyfill for PHP 8's `Stringable` interface, thus allowing consumers to type-hint against that in 2.x and future versions.

I tried other implementations to avoid this BC-break but ultimately found them to be too kludgy and not developer-friendly.

## Future Scope

This implementation does not include the config override functionality.  I've got it working locally but we'll need to address https://github.com/thephpleague/commonmark/issues/442#issuecomment-615895227 first.  We'll save that config override stuff for a near-future PR :wink:

### DIFF
--- a/src/Input/MarkdownInput.php
+++ b/src/Input/MarkdownInput.php
@@ -38,13 +38,21 @@ class MarkdownInput implements MarkdownInputInterface
      */
     private $lineCount;
 
-    public function __construct(string $content)
+    /**
+     * @var int
+     *
+     * @psalm-readonly
+     */
+    private $lineOffset = 0;
+
+    public function __construct(string $content, int $lineOffset = 0)
     {
         if (! \mb_check_encoding($content, 'UTF-8')) {
             throw new UnexpectedEncodingException('Unexpected encoding - UTF-8 or ASCII was expected');
         }
 
-        $this->content = $content;
+        $this->content    = $content;
+        $this->lineOffset = $lineOffset;
     }
 
     public function getContent(): string
@@ -60,8 +68,8 @@ class MarkdownInput implements MarkdownInputInterface
         $this->splitLinesIfNeeded();
 
         /** @psalm-suppress PossiblyNullIterator */
-        foreach ($this->lines as $lineNumber => $line) {
-            yield $lineNumber => $line;
+        foreach ($this->lines as $i => $line) {
+            yield $this->lineOffset + $i + 1 => $line;
         }
     }
 
--- a/src/Input/MarkdownInput.php
+++ b/src/Input/MarkdownInput.php
@@ -15,7 +15,7 @@ namespace League\CommonMark\Input;
 
 use League\CommonMark\Exception\UnexpectedEncodingException;
 
-final class MarkdownInput implements MarkdownInputInterface
+class MarkdownInput implements MarkdownInputInterface
 {
     /**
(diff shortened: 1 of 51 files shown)


colinodell:
TODO: Rebase this onto `latest`; resolve merge conflicts

colinodell:
@markcarver If you have any bandwidth to give this a quick sanity check I'd appreciate it!

src/Input/MarkdownInput.php
```diff
@@ -38,13 +38,21 @@ final class MarkdownInput implements MarkdownInputInterface
...
+
+    public function __construct(string $content, int $lineOffset = 0)
     {
         if (! \mb_check_encoding($content, 'UTF-8')) {
             throw new UnexpectedEncodingException('Unexpected encoding - UTF-8 or ASCII was expected');
         }
 
-        $this->content = $content;
+        $this->content    = $content;
+        $this->lineOffset = $lineOffset;
```
markcarver, line 55: There's no `getLineOffset` method which could be quite useful on its own (accurate debug tracing)

markcarver (approved):
Aside from the one missing method (which can be done as a follow-up), this PR LGTM.

src/Input/MarkdownInput.php
colinodell: Lines only get numbered when iterating over them via `getLines()` which does utilize that variable.

Issue #531, opened 2020-08-14
Heading permalinks don't work if a document has the same title multiple times


onigoetz:
**Version(s) affected**: 1.5.3

**Description**  

When using the Heading permalinks, if a document has the same title multiple times, it creates the same link, which won't work

**How to reproduce**  

```php
<?php

require "vendor/autoload.php";

use League\CommonMark\CommonMarkConverter;
use League\CommonMark\Environment;
use League\CommonMark\Extension\HeadingPermalink\HeadingPermalinkExtension;

// Obtain a pre-configured Environment with all the CommonMark parsers/renderers ready-to-go
$environment = Environment::createCommonMarkEnvironment();

// Add this extension
$environment->addExtension(new HeadingPermalinkExtension());

// Instantiate the converter engine and start converting some Markdown!
$converter = new CommonMarkConverter([], $environment);
echo $converter->convertToHtml("# Hello World!\n## Title\n## Title");
```

Outputs
```
<h1><a id="user-content-hello-world" href="#hello-world" name="hello-world" class="heading-permalink" aria-hidden="true" title="Permalink">¶</a>Hello World!</h1>
<h2><a id="user-content-title" href="#title" name="title" class="heading-permalink" aria-hidden="true" title="Permalink">¶</a>Title</h2>
<h2><a id="user-content-title" href="#title" name="title" class="heading-permalink" aria-hidden="true" title="Permalink">¶</a>Title</h2>
```

The second `#title` should be `#title-1` or `#title-2`

I checked on babelmark; https://babelmark.github.io/?text=%23+title%0A%23%23+title and it's not clear what the take is on this.

I would love to use your implementation and throw away the ... thing ... I made for Daux : https://github.com/dauxio/daux.io/blob/master/libs/Format/HTML/ContentTypes/Markdown/TOC/Processor.php#L71

It's an all-in-one solution to add IDs to headers and create a Table of Contents.
The separation you made between headings and TOC seems way cleaner and I'll adopt it once it can do multiple titles as well.

P.S. Thanks for maintaining this parser, it's a really neat implementation and easy to use :)

colinodell:
> I checked on babelmark; https://babelmark.github.io/?text=%23+title%0A%23%23+title and it's not clear what the take is on this.

Because this is a GFM feature, we should follow the example of the official GFM parser which does what you propose.

onigoetz:
I feel the way you fixed it makes it mandatory to create a new instance of the parser for each document that will be parsed.

[...]

PR #534, opened 2020-08-15
Enforce slug uniqueness per-document

colinodell:
Fixes https://github.com/thephpleague/commonmark/issues/531#issuecomment-674390724

@onigoetz Would you mind reviewing this approach (which is based on what you've been using)?


onigoetz:
Seems, to work fine, looks good to me

PR #540, opened 2020-09-10
Fix non-static methods called as static '::' methods

Ayesh:
Fixes a couple static calls in `TableParser` class that used static method call pattern (`::`) on non-static calls ('->`).

Thank you.


GrahamCampbell:
I think the bug is that the method is not static, not that it is called wrong? The method should instead be changed to be static.

Ayesh:
Hi @GrahamCampbell - Thanks for the review. 
`parseCell` is a private method and `TableParser` is declared final, but `parseCell` method uses instance properties (`$this->columns`) so I propose that we change the call to a non-static call.

GrahamCampbell:
Oh, yeh.

GrahamCampbell (approved):

colinodell:
Thanks for the fix!

PR #542, opened 2020-09-10
Fix LSP incompatibility in \League\CommonMark\Environment\Environment::getConfig

Ayesh:
Fixes a missing scalar type in the first parameter of `\League\CommonMark\Environment\Environment::getConfig()`. This method does not follow [LSP](https://php.watch/articles/php-lsp) and errors out.


colinodell:
Thanks for catching this!

Issue #421, opened 2020-04-04
Feature Request: Emoji Parser


colinodell:
See https://github.com/thephpleague/commonmark-extras/issues/19

iksaku:
Which would be a good way to keep emojis’ Unicode/image mapping?

I’m currently thinking of two possible ways, considering that we don’t want to hook up to  Github for this:
- Having the map hard coded in class constants
- Have them available in config options for merge/override (I kinda prefer this one)

Any other ideas? Maybe I could send a PR with my personal implementation (which uses config options ATM)

glensc:
Maybe have some third party composer package for the Unicode mappings. it could be then updated by its own schedule, from GitHub or whatever sources, and provide clean API to access the mapping.

Also, I'm pretty sure it should be done as Unicode output, if someone wants to extend to use images, they can do that top of the conversion, probably just on client-side:

```html
<span class="emoji" data-emoji-codepoint="1F60A" data-emoji-shortcode="blush">&#x1F60A;</span>
```

I think gitlab at some point struggle with this, and they emit in HTML, and replace with images only if browser doesn't support colored emojis. need some digging in their issue or merge requests or maybe they even wrote blog.

glensc:
Seems github uses emojione:
- https://github.com/bonusly/gemojione

which php variant is:
- https://packagist.org/packages/emojione/emojione

but that's superseded by:
- https://github.com/joypixels/emoji-toolkit

colinodell:
Dumping some of my initial thoughts here:

 - The parser should look for both Unicode character and `:+1:`-type syntax create an `Emoji` inline element in the AST
 - The renderer will be responsible for determining whether to render those as images, Unicode characters, or something like @glensc's `<span>` examples. Not 100% sure on the image part, but if so, we should make it easy for people to plug in their own image set or provide a sane default.
 - We need the ability to map back-and-forth between Unicode characters and emoji names so that both `:+1:` and `👍` end up with similar (if not identical) `Emoji` element representations
 - What about support for custom emojis? For example, I might want `:partyparrot:`.  How can we support that? What happens if the user chooses to render the `Emoji` elements as Unicode characters - what should happen to this emoji here?

iksaku:
### About Initial Short Code availability 
[...]

PR #546, opened 2020-09-17
WIP: Emoji extension

markhalliwell:
This is a WIP for #421, an emoji extension implementation using https://github.com/unicorn-fail/emoji

~~I haven't yet actually tried this to see if it works as expected, this work is based solely on IDE reliance on API implementation.~~

This, currently, includes a single test as a POC

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         "scrutinizer/ocular": "^1.5",
         "symfony/finder": "^5.1",
         "symfony/yaml": "^2.3 | ^3.0 | ^4.0 | ^5.0",
+        "unicorn-fail/emoji": "1.0.x-dev",
         "unleashedtech/php-coding-standard": "^2.5",
         "vimeo/psalm": "^3.14"
     },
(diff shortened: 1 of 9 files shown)


colinodell:
This is looking great so far!

giberti:
I took a stab at implementing this for a project I've been working on. It's functional, but lacks a full 1:1 mapping of :octocat: emoji aliases to Unicode CLDR names. I felt preserving the official names was important for internationalization.

https://github.com/giberti/commonmark-emoji-extension

markcarver:
The actual parsing/rendering of emojis should be handled by a separate library. That is why we're creating https://github.com/unicorn-fail/emoji (will be migrated to `thephpleague` namespace after its documentation is finalized).

This issue will merely be an extension that implements that.

giberti:
Yeah, I agree! My version of building the data files is no where as far along as yours; and doesn't include the generation script because it's not pretty. 😆
 
https://github.com/giberti/emoji-data

giberti:
Looking forward to switching over to this one soon.

base-zero:
Is there any update on this PR and in turn an Emoji extension for commonmark ? This is the one feature that is stopping us for moving over to commonmark fully in our project.

PR #550, opened 2020-09-26
Allow more than one character to be used as a symbol for MentionExtension

### LINKED ISSUE #514: Allow more than one character to be used as a symbol for MentionExtension.

colinodell:
Thanks to #549 we're able to implement this for the upcoming 2.0 release

Resolves #514

### DIFF
--- a/src/Extension/Mention/MentionParser.php
+++ b/src/Extension/Mention/MentionParser.php
@@ -70,7 +70,7 @@ final class MentionParser implements InlineParserInterface
         $previousState = $cursor->saveState();
 
         // Advance past the symbol to keep parsing simpler
-        $cursor->advance();
+        $cursor->advanceBy(\mb_strlen($this->symbol));
 
         // Parse the identifier
         $identifier = $cursor->match($this->mentionRegex);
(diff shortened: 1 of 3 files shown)


src/Extension/Mention/MentionParser.php
```diff
@@ -70,7 +70,7 @@ public function parse(string $match, InlineParserContext $inlineContext): bool
         $previousState = $cursor->saveState();
 
         // Advance past the symbol to keep parsing simpler
-        $cursor->advance();
+        $cursor->advanceBy(\mb_strlen($this->symbol));
 
         // Parse the identifier
         $identifier = $cursor->match($this->mentionRegex);
```
markcarver, line 76: I feel like both the symbol and mention regex should be combined into a single regex that uses parenthesis to group the individual results. Maybe something like:

```php
        return InlineParserMatch::join(
          InlineParserMatch::string($this->symbol),
          InlineParserMatch::regex($this->mentionRegex),
        );
```

Unfortunately, this would also mean that `string $match` would have to really be `array $matches`.

Also, is there a reason we can't add both of these to `InlineParserContext`, i.e. `$inlineContext->getMatch(): string` and `$inlineContext->getMatches(): array`; the former just returnining the first item of the latter?

Then we could something like the following here:

```php
        // Extract the symbol and identifier from the matches.
        [$symbol, $identifier] = $inlineContext->getMatches();

        // Advance past the symbol to keep parsing simpler
        $cursor->advanceBy(\mb_strlen($symbol));
```

colinodell: Great idea!  I've incorporated this into #560.

Issue #475, opened 2020-05-14
[2.0] Keep high-level API as similar to 1.x as possible


colinodell:
Many users only care about adding extensions, setting configs, and converting Markdown.  Let's try to keep those APIs as similar as possible, even if a lot of the "internal" bits have BC-breaks.

This will require a review of those APIs and possibly a new 1.x minor release to denote deprecations to end-users.

(For example, we could add class aliases in 1.x so that users can be compatible with both that newer 1.x release and 2.x.  This may not be possible or feasible in all cases though - needs more research)

A sample, incomplete list of such things:

 - `CommonMarkConverter`
 - `ExtensionInterface`
 - Things in the new `Environment` sub-namespace

colinodell:
Let's try to get as much forward-compatibility and deprecation notices into 1.5 as possible :wink:

colinodell:
# Current status (WIP)

(This will be updated as I review the classes over the comings days and weeks)

## ✔ CommonMarkConverter

In 2.x, the base `Converter` class is gone, but all the other important bits should be compatible:

| Feature | Similar Enough? | Notes |
| --- | --- | --- |
| Base interface | ✔ | Both implement `MarkdownConverterInterface` |
| `__construct()` | ✔ | The `EnvironmentInterface` is in a different namespace, but the argument position is the same |
| `getEnvironment()` | ✔ | The `EnvironmentInterface` is in a different namespace |
| `convertToHtml()` | ✔ | Identical |
| `__invoke()` | ✔ | Identical |
| `public const VERSION` | ✔ | Identical  |

## ConfigurableEnvironmentInterface

A couple of methods have changed, and this lives in a different namespace now, but the method to add extensions is the same:

| Feature | Similar Enough? | Notes |
| --- | --- | --- |
| `addExtension()` | ✔ | Identical |

**TODO:** Update this based on #436 and #447

colinodell:
**TL;DR:** After discussing with Mark Carver I've decided that, in the best interest of this project and its users, no further effort will be put into forward-compatibility for 1.5; instead, we'll focus on minimizing BC-breaks in 2.x and making that version the best it can be.  See the longer thread below for why.

-----

Generally speaking, there are three types of people who use this library:

1. People who instantiate a `CommonMarkConverter`, maybe add some extensions, and convert Markdown
2. People who integrate this library with other projects/frameworks and who use some of the other high-level APIs, like extensions
3. People who create custom parsers, event listeners, etc.

[...]

PR #551, opened 2020-09-26
Rename "symbol" option to "prefix"

colinodell:
Once #550 is merged we should also consider renaming "symbol" (which implies a single character) to "prefix" for v2.0 onwards.

Because this would be a breaking change to configuration inputs (against the spirit of #475) we could release a new patch version for 1.x where both `symbol` and `prefix` are allowed as configuration keys but a deprecation notice is raised if `symbol` is used.

WDYT @markcarver?

(Note that this PR is based directly on the commit history present in #550)


markcarver (commented):
Honestly, at this point, the symbol/prefix is now rather redundant. Can't we just combine it with the regex since that whole thing could be passed as the match definition? Assuming, of course, we had some sort of regex group support and were able to extract [an array of matches instead of assuming it'll just return a single string](https://github.com/thephpleague/commonmark/pull/550#discussion_r498107656).

Issue #479, opened 2020-05-16
Add description list support like PHP Markdown Extra


ptmkenny:
**Description**  

Please add support for description lists (definition lists) as implemented by [PHP Markdown Extra ](https://michelf.ca/projects/php-markdown/extra/#def-list) and also supported by [Pandoc](https://pandoc.org/MANUAL.html#definition-lists), [MultiMarkdown](http://fletcher.github.io/MultiMarkdown-4/definitionlists), [Kramdown](https://kramdown.gettalong.org/syntax.html#definition-lists), [markdown-it](https://github.com/markdown-it/markdown-it-deflist), and [Python Markdown](https://python-markdown.github.io/extensions/definition_lists/).

This is not officially part of the CommonMark spec yet, but jgm [has commented favorably on this implementation](
https://talk.commonmark.org/t/description-list/289/14) and it is widely implemented.

**Example**

Input:
  
```
Apple
: This may refer to a fruit, a record company, or a phone maker.
```

Evaluates to:

```
<dl>
<dt>Apple</dt>
<dd>This may refer to a fruit, a record company, or a phone maker.</dd>
```

colinodell:
Thanks for the idea, and for providing the references showing support for this syntax in other parsers and by jgm!

colinodell:
This is going to be tricky to implement in 1.x. I'm going to punt this to 2.0 as the new parser system should more-easily support the kind of nested parsing we'll need to properly implement this.

aboks:
I'm just seeing this issue now, but I've created a plugin to support this syntax with league/commonmark 1.x that you could use in the meantime: https://github.com/Moxio/commonmark-ext-definition-list. I indeed noticed that the parser is currently not really suited for syntax like this (mainly because you only know at the line with the `:` that the line before was part of the list), but it seems to work fine for now.

@colinodell If you ever want to adopt that plugin (or parts of it) into core, feel free to reach out.

PR #552, opened 2020-09-29
Add Definition List ext to community extensions

aboks:


markcarver:
Appreciate the new extension, good work! I'd rather the goal here be to add/merge this new extension directly into this project as that was the intention of #479.

Issue #553, opened 2020-09-30
Improve documentation


colinodell:
I'm not the greatest at writing documentation. Sometimes I don't fully proofread my work, and sometimes I assume that my readers already understand concepts I might not have explained sufficiently.  I'm also terrible at writing good CSS.

If you'd like to help me improve the documentation I'd greatly appreciate it!

## Who Should Help?

In the spirit of Hacktoberfest, contributions are welcome from anyone and everyone, especially if you are new to contributing to open-source projects!

## Tips

You can view the documentation online at https://commonmark.thephpleague.com/.  It is stored in this repository at https://github.com/thephpleague/commonmark/tree/latest/docs.

You can use Jekyll to spin up a local copy of the docs to test and preview your changes.  Assuming you already have [Ruby](https://www.ruby-lang.org/en/documentation/installation/), and [Bundler](https://bundler.io/) installed, simply run:

```bash
cd docs # Switch to the directory with the documentation site files
bundle install # Install Jekyll and other dependencies

bundle exec jekyll serve # Build the site
```

That last command will give you a link to preview the site in your browser.  Jekyll will watch for any changes you make to the site files, meaning you can simply reload your browser to see the changes in near-real-time.

## Acceptance Guidelines

Please keep the following in mind when suggesting changes to ensure your PR has a good chance of being quickly accepted:

**Examples of good changes:**

 - Fixing typos and other grammar issues
 - Fixing any broken code examples
 - Standardizing formatting between different pages
 - Expanding existing sections which don't have enough information
 - Improving my terrible CSS and fixing styling issues

**Please avoid:**

 - Submitting pull requests to any branch other than `latest`
 - Making edits that do not contribute in a meaningful way (for example, changing words like `color` to `colour`)
 - Editing documentation for older versions without also applying the same changes to newer versions
 - Sending one giant pull request with TONS of changes - smaller, more-focused PRs are easier and quicker to review

## Questions?

If you have any questions or want to suggest larger changes, please feel free to do so in this thread :)

**Thanks for contributing!**

Rushikesh-ganesh:
i would like to contribute @colinodell

PR #554, opened 2020-10-02
Code block consistency

pdelre:
Made a pass on consistent use of code block formatting through the `1.5` docs. I made a few assumptions in applying the changes:

- Backticks are preferred over indentation and tildes
  - 42 `php` code blocks used backticks (`rg -U '```php\s+' | wc -l`) vs 35 `php` code blocks using tildes (`rg -U '~~~php\s+' | wc -l`)
- `php` code blocks should not include PHP open tags (`<?php`)
  - 88 `php` code blocks did not include an open tag (`rg -U '(```|~~~)php\s+[^<]' | wc -l`) vs 66 `php` code blocks included an open tag (`rg -U '(```|~~~)php\s+[<]' | wc -l`)
- Code blocks should specify their language
- `markdown` is preferred over `md` for it's explicitness

I was unable to covers two points made in #553:

> - Editing documentation for older versions without also applying the same changes to newer versions
> - Sending one giant pull request with TONS of changes - smaller, more-focused PRs are easier and quicker to review

I did not update any other version of the documentation (violating the first point) and my many small changes still resulted in a large, but atomic, PR (violating the second point). @colinodell if this PR looks good, I'll expand my efforts. In that, would you prefer PRs for each version (but limited to code block consistency) or one large PR across all versions (but limited to code block consistency)? Asking in good faith for as I am participating in Hacktoberfest, I know the event brings a spike in load for project maintainers which includes filtering spam PRs, which is not my intent.

### DIFF
--- a/docs/1.5/basic-usage.md
+++ b/docs/1.5/basic-usage.md
@@ -10,22 +10,18 @@ Basic Usage
 
 The `CommonMarkConverter` class provides a simple wrapper for converting Markdown to HTML:
 
-~~~php
-<?php
-
+```php
 use League\CommonMark\CommonMarkConverter;
 
 $converter = new CommonMarkConverter();
 echo $converter->convertToHtml('# Hello World!');
 
 // <h1>Hello World!</h1>
-~~~
+```
 
 Or if you want Github-Flavored Markdown:
 
 ```php
-<?php
-
 use League\CommonMark\GithubFlavoredMarkdownConverter;
 
 $converter = new GithubFlavoredMarkdownConverter();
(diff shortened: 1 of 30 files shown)


colinodell (approved):
@pdelre these changes look great!  I agree with the choices you've made.  Nicely done :)

Because versions 1.4 and older are no longer supported, we don't necessarily have to update them.  I mostly only care about the documentation for 1.5 and 2.0.  However, if you'd like some "extra credit" to make similar changes in those older versions, I will gladly accept them.  One large PR would be fine with me for that.

PR #555, opened 2020-10-02
Use alternative ascii characters in string template illustration

pdelre:
When reviewing the docs, I was slowed by trying to grok the ascii based illustration of String-Based URL Templates for the Mention extension. It took a moment to understand that the `Identifier` underline wasn't offset, but pointing to two different areas (`colinodell`) and not highlighting the section above it `dell" => https://www.twitter.com/colin`.

image

I've switched it to use some of the additional ascii characters to highlight that the `colinodell` to the direct right of the `@` symbol was the identifier that was carried over to the value on the right. I also used an arrow instead of carot and put `Symbol` on the same line as `Identifier`, giving them equal weight.

image

RE: #553

### DIFF
--- a/docs/1.5/extensions/mentions.md
+++ b/docs/1.5/extensions/mentions.md
@@ -74,9 +74,9 @@ URL templates are perfect for situations where the identifier is inserted direct
 
 ```
 "@colinodell" => https://www.twitter.com/colinodell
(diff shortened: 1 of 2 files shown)


colinodell (approved):
Thank you so much for your contribution!  This is perfect :tada:

PR #556, opened 2020-10-03
Update the link for autoloading with composer

martijnengler:
The installation instructions mention enabling "autoloading" for
composer-installed packages.

The composer project has moved this documentation from the intro-page,
to the "basic usage"-page, which broke the link in our docs.

This commit fixes it for all versions of the documentation.


colinodell (approved):
Thank you!

PR #557, opened 2020-10-03
Autoloader instructions

martijnengler:
If someone is new with composer (which is entirely possible) it might be confusing to get a "class not found" error when copy/pasting the example code.

This commit adds including the autoloader to the examples for basic usage, so they can be literally copy/pasted in a text editor and will work right away.

This can make it easier to get started for beginners.

### DIFF
--- a/docs/0.19/basic-usage.md
+++ b/docs/0.19/basic-usage.md
@@ -11,6 +11,8 @@ The `CommonMarkConverter` class provides a simple wrapper for converting CommonM
 ~~~php
 <?php
 
+require __DIR__ . '/vendor/autoload.php';
+
 use League\CommonMark\CommonMarkConverter;
 
 $converter = new CommonMarkConverter();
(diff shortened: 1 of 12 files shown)


docs/0.19/basic-usage.md
```diff
@@ -11,6 +11,8 @@ The `CommonMarkConverter` class provides a simple wrapper for converting CommonM
 ~~~php
 <?php
 
+require __DIR__ . '/vendor/autoload.php';
```
GrahamCampbell, line 14: require_once is what we probably want here?

GrahamCampbell (changes requested):

docs/0.19/basic-usage.md
martijnengler: It's funny you should say that, because that's what I wrote at first as well!

But I noticed composer itself uses just `require` (see https://getcomposer.org/doc/01-basic-usage.md#autoloading), so I decided to follow their convention.

Happy to change it though.

colinodell: While I personally prefer `require_once`, I like the idea of following Composer's convention in our examples so beginners can follow along with similar instructions/examples in both places.

PR #560, opened 2020-10-03
Refine inline parsing

colinodell:
This takes the changes from #550 and #551 to the next level :wink: 

Based on @markcarver's feedback, we're removing the `string $match` parameter from `InlineParserInterface::parse()` and instead exposing the matches via the context.  This also now includes sub-matches from regex capturing groups.

### DIFF
--- a/tests/functional/Extension/Mention/MentionExtensionTest.php
+++ b/tests/functional/Extension/Mention/MentionExtensionTest.php
@@ -15,6 +15,7 @@ namespace League\CommonMark\Tests\Functional\Extension\Mention;
 
 use League\CommonMark\CommonMarkConverter;
 use League\CommonMark\Environment\Environment;
+use League\CommonMark\Exception\InvalidOptionException;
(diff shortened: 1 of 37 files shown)


tests/functional/Extension/Mention/MentionExtensionTest.php
```diff
@@ -58,8 +59,8 @@ public function testConfigStringGenerator(): void
         $environment->setConfig([
             'mentions' => [
                 'github_handle' => [
-                    'symbol'    => '@',
-                    'regex'     => '/^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}(?!\w)/',
+                    'prefix'    => '@',
+                    'regex'     => '[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}(?!\w)',
```
markcarver, line 63: Why are we still including a `prefix`? Shouldn't this just be part of the `regex` now?

PR #562, opened 2020-10-03
Fix styles in table headings

tobiasthaden:
This PR adds inheritance to all possible direct children of a table header in the docs.

Currently, the `th` styles are bypassed by the styles of the child element. 
E.g. https://commonmark.thephpleague.com/1.5/upgrading/


colinodell (approved):
Looks good to me.  Thanks for contributing! :tada:

PR #568, opened 2020-10-06
Update 2.0 docs to reflect symbol -> prefix change

markcarver:


docs/2.0/extensions/mentions.md
```diff
@@ -35,25 +35,25 @@ $config = [
         // Sample Input:  `@colinodell`
         // Sample Output: `<a href="https://www.github.com/colinodell">@colinodell</a>`
         'github_handle' => [
-            'symbol'    => '@',
+            'prefix'    => '@',
             'regex'     => '/^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}(?!\w)/',
```
colinodell, line 39: Good catch!  We also need to remove the starting `/^` and trailing `/` from all the regexes.

markcarver: Ah, right... done.

markcarver: This almost makes me wonder if it should be called `pattern` now instead of `regex`?

colinodell: I like that :)

markcarver: That can be a follow-up. I don't have time to include that in this PR.

colinodell: I've created https://github.com/thephpleague/commonmark/issues/569 to track that

PR #570, opened 2020-10-09
Code block consistency - Legacy versions

pdelre:
Follow up to https://github.com/thephpleague/commonmark/pull/554#pullrequestreview-501455569. This follows the same decisions in that PR and applies it backward to previous version's docs. I separated this out as I used more automated processes for the edits and this PR is less likely to have conflicts than the `2.0` docs which are under active development.

RE: #553


colinodell (approved):
Looks good to me. Thank you so much!

PR #573, opened 2020-10-09
Add “name” property to mentions so implementers can identify which mention is which

markcarver:
While working on https://www.drupal.org/project/markdown/issues/3160927, I suddenly realized that there's no "easy" way to determine which mention is being used by the generator callback we're using, thus I cannot load that specific mention's configuration (has much more than the normal mention due to CMS integrations).

Since the changes to these classes are mostly internal, I figured it wouldn't hurt to add a new parameter that names the mention (which is automatically pulled from the mention configuration as the array key).

### DIFF
--- a/src/Extension/Mention/Mention.php
+++ b/src/Extension/Mention/Mention.php
@@ -22,13 +22,17 @@ use League\CommonMark\Node\Inline\Text;
 class Mention extends Link
 {
     /** @var string */
+    private $name;
+
+    /** @var string */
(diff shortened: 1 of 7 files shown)


markcarver:
P.S. do you think maybe this could be backported?

colinodell (approved):

colinodell:
> P.S. do you think maybe this could be backported?

Unfortunately no, because this change would break backward-compatibility.

PR #574, opened 2020-10-10
Fix 2.0 mention docs to use pattern instead of regex

markcarver:
Oops, we missed this


colinodell (approved):
Good catch!  There are too many docs to keep track of :upside_down_face:

PR #572, opened 2020-10-09
Code block consistency - Upcoming version

pdelre:
Follow up to https://github.com/thephpleague/commonmark/pull/554#pullrequestreview-501455569. This follows the same decisions in that PR and applies it to the `2.0/` documents. I separated this out as it is more likely to contain conflicts as 2.0 is under active development. Furthermore, these changes were applied in a more hands on manner and manually checked throughout the rendered site.

RE: #553

PR #575, opened 2020-10-10
Add instructions on how to install dev version

martijnengler:
When version 2.0 goes stable, it will be as easy as is currently
documented:

```bash
composer require "league/commonmark:^2.0"
```

Right now however, there's a good chance people have their composer
settings set to have a minimum stability of "stable".[1]
This will probably result in errors when trying to run this command.

I wasn't sure what the best way forward is for here: leave in the old,
but make a note of it? Make temporary instructions for dev?

For now I went with a combination: leave the current way (with just
`composer require`), but add some instructions for the current dev
version.

When this version goes stable the extra instructions can easily be
deleted from the document.

[1] This is the issue I ran into myself when I tried to play around with 
the 2.0 version. This PR is my way of paying it forward or something. 😉

### DIFF
--- a/docs/2.0/installation.md
+++ b/docs/2.0/installation.md
@@ -34,7 +34,7 @@ Put the following code in it and save:
 
 When the file is set up, run the following command and you are done.
 
-```sh
+```bash
 composer install
 ```
 
--- a/docs/2.0/installation.md
+++ b/docs/2.0/installation.md
@@ -8,10 +8,34 @@ description: Instructions on how to install the league/commonmark library
 
 The recommended installation method is via Composer.
 
-In your project root just run:
-
 ```bash
-composer require league/commonmark:^2.0
+composer require "league/commonmark:^2.0"
+```
+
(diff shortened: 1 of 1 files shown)


docs/2.0/installation.md
```diff
@@ -8,10 +8,34 @@ description: Instructions on how to install the league/commonmark library
...
+	"require-dev": {
+		"league/commonmark": "^2.0@dev",
+		"dflydev/dot-access-data": "^3.0@dev"
+	}
+}
+```
+
+When the file is set up, run the following command and you are done.
+
+```sh
```
pdelre, line 37: I've been doing some consistency reviews of `docs/` and `bash` is preferred over `sh` for code block language selection.

```suggestion
```bash
```

martijnengler: Yep, saw your work in #572 , didn't mean to undo that. Thanks for noticing/commenting, fixed it! :)

colinodell (approved):
This works for me!  Thank you!

PR #577, opened 2020-10-16
Documentation changes

miken32:
A few issues fixed up per #553. Missing punctuation, clarifying some example code, and adding output samples to some MD syntax.


colinodell (approved):
Thank you for these edits!  Hopefully they don't conflict with #578 but we'll deal with that separately if so.

I appreciate the contribution -  Happy Hacktoberfest! 🎃

PR #578, opened 2020-10-16
Add markdownlint

### LINKED ISSUE #576: RFC: Add markdown linting for docs/
**Description**  

#553 mentions a goal _"Standardizing formatting between different pages"_. To that end, the project could adopt a markdown linter for their docs. Adopting a linter will require various updates to the repo, its contributing documentation, and it's CI if it would like to be enforced.

1. Would the maintainers like an addition of a linter that checks the `docs/`, even with the (hopefully minimal) added burden of project configuration and maintenance changes?

I've generally used JS's [DavidAnson/markdownlint](https://github.com/DavidAnson/markdownlint) as it integrates nicely with VSCode, but it was ported from ruby's [markdownlint/markdownlint](https://github.com/markdownlint/markdownlint), which aligns with the project's usage of Jekyll. The JS version has the added advantage of being used by GitHub's [Super-Linter Action](https://github.com/github/super-linter), further reducing the CI maintenance cost.  Unfortunately the two implementations are not compatible and cannot share configuration files.

2. Would the maintainers prefer `DavidAnson/markdownlint` or `markdownlint/markdownlint` as the chosen implementation?

**VS Code Example**  

image

pdelre:
Closes #576 

This adds `.markdownlint.yaml` for [DavidAnson/markdownlint](https://github.com/DavidAnson/markdownlint/) and adds a [Super-Linter GitHub Action](https://github.com/github/super-linter) to check the `docs/1.5/` and `docs/2.0/` directories.

This also fixes any discovered issues in `docs/1.5/` and `docs/2.0/`.

RE: #553


.github/workflows/docs-lint.yml
```diff
@@ -0,0 +1,24 @@
+---
+name: Lint Docs
```
pdelre, line 2: First time using GitHub Actions, so special attention is needed here.

colinodell: This looks good to me! Nicely done :)

.github/workflows/docs-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+
+      - uses: github/super-linter@v3
+        env:
+          DEFAULT_BRANCH: latest
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          LINTER_RULES_PATH: '.' # From https://github.com/github/super-linter/pull/859#issuecomment-709625090
+
+          VALIDATE_MARKDOWN: true
+          MARKDOWN_CONFIG_FILE: docs/.markdownlint.yml
+          FILTER_REGEX_INCLUDE: '^docs/(1\.5|2\.0)/.*'
```
colinodell, line 24: Can we also lint the Markdown files that live outside the version-specific directories?  If that means we also have to lint the older versions too I'd be okay with that.

pdelre: Yes, I'll pick that up next week. Do you want it on only docs or the README and root project markdown files?

colinodell: Great question! I think just the docs - I don't mind if the README, CHANGELOG, etc. are slightly different.  And we don't want to lint any of the Markdown files used in our tests.

No rush on my end.  I appreciate your help here!

pdelre: @colinodell All of `docs/` is now linted and all errors have been resolved.

colinodell (approved):
This looks great! Thank you so much!

PR #582, opened 2020-10-27
Fix mentions when they appear after a non-word character such as \n or ,

jean-gui:
(Copied from https://github.com/thephpleague/commonmark/pull/581)

**Version(s) affected**: 1.5.6

**Description**  
If a mention immediately follows a new line or any other non-word character it doesn't get converted to a link.

**How to reproduce**  
```php
$environment = Environment::createGFMEnvironment();
$environment->addExtension(new MentionExtension());

$converter = new GithubFlavoredMarkdownConverter(
    [
        'mentions' => [
            'github_handle'  => [
                'symbol'    => '@',
                'regex'     => '/^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}(?!\w)/',
                'generator' => 'https://github.com/%s',
            ]
        ],
    ],
    $environment
);

echo $converter->convertToHtml("@test\n@test");
```
Results in:
```html
<p><a href="https://github.com/test">@test</a>
@test</p>
```
instead of
```html
<p><a href="https://github.com/test">@test</a>
<a href="https://github.com/test">@test</a></p>
```
**Fix**
This PR replaces the test for a space with one for any non-word character.
This seems to be what GitHub does, as `test,@jean-gui` gets converted to `test<a href="...">@jean-gui</a>`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
 
 ## [Unreleased][unreleased]
 
+### Fixed
+
+ - Fixed mentions not being parsed when appearing after non-word characters (#582)
+
 ## [1.5.6] - 2020-10-17
 
 ### Changed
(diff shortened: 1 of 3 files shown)


colinodell:
Thanks for the fix!

PR #583, opened 2020-10-29
Docs: Improve readabillity on large screens

tobiasthaden:
This PR adds a `max-width` on paragraphs, to limit the line length and improve the readabillity on larger screens.


colinodell (approved):
Great idea! Thanks for the improvement :)

PR #608, opened 2021-04-08
fix documentation casing error by replace DisallowedRawHTML with Disa…

### LINKED ISSUE #607: Naming of DisallowedRawHTML inconsistent
**Version(s) affected**: 1.5

**Description**  

According to the docs: https://commonmark.thephpleague.com/1.5/extensions/disallowed-raw-html/
You use the extension with in Usage
> use League\CommonMark\Extension\DisallowedRawHTML\DisallowedRawHtmlExtension;

But, this does not work, since the extension / class is actually called DisallowedRawHtml (HTML vs. Html).

Something similar discussed in #599 .
But, that was not enough.

**Possible Solution**  

The code examples / documentation should be fixed, I guess.

Would you mind if I fix this bug?

kinpoko:
…llowedRawHtml (Html instead of HTML)

Replace DisallowedRawHTML with DisallowedRawHtml (Html instead of HTML) in documentation
fix #607

### DIFF
--- a/docs/1.3/extensions/disallowed-raw-html.md
+++ b/docs/1.3/extensions/disallowed-raw-html.md
@@ -43,7 +43,7 @@ Configure your `Environment` as usual and simply add the `DisallowedRawHtmlExten
 ```php
 use League\CommonMark\CommonMarkConverter;
 use League\CommonMark\Environment;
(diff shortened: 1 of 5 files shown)


markhalliwell (approved):
LGTM

colinodell:
Thanks!

PR #609, opened 2021-04-22
Use GitHub instead of Github

Sea-n:
The brand name should be Git**H**ub.

This PR leaves `GithubFlavoredMarkdown` classes as is, this should be a non-breaking change.


colinodell (approved):
Thank you!

PR #612, opened 2021-04-30
ci: fix code coverage

bytestream:
Code coverage is failing in CI but the job still passes due to `continue-on-error: true`

```
In RepositoryIntrospector.php line 39:
                                                                               
  Failed to retrieve commit parents. If you use a shallow git checkout, pleas  
  e checkout at least a depth of one.                                          
                                                                               

code-coverage:upload [--api-url API-URL] [--repository REPOSITORY] [--revision REVISION] [--format FORMAT] [--parent PARENT] [--] <coverage-file>
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,6 +41,8 @@ jobs:
 
         steps:
             - uses: actions/checkout@v2
+              with:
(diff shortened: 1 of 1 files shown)


colinodell:
Thanks for the fix!

Issue #605, opened 2021-03-30
Question: How to add classes on specific node ?


armetiz:
## Question

I have a common markdown content, without any specific attributes.
When rendering markdown to HTML, I want to see some CSS Classes on HTML nodes.

Let's say : 
* All H1 nodes : `.title-main`
* All Table nodes : `.table`
* All Table Row nodes : `.table .table-row`

Is this possible without creating custom renderer ?

It could be awesome to be able to configure using the environment : 

```
$config = [
  'html_classes' => [
    'h1' => '.title-main',
    'table' => '.table',
  ],
];
```

The use case is : I'm using a bought theme that request some css class. I don't want to change markdown original content, and I don't want to change theme.

Thank a lot for your awesome job on this package ! 

Regards,

colinodell:
You can create an event listener the handles the `DocumentParsedEvent`, iterates through the parsed nodes, and applies the custom classes as needed.  See https://commonmark.thephpleague.com/1.5/customization/event-dispatcher/#example for an example of this.

In your case, that code would look something like this:

```php
use League\CommonMark\EnvironmentInterface;
use League\CommonMark\Event\DocumentParsedEvent;
// TODO: add other missing "use" statements here

class AddCustomCssClassesProcessor
{
    private $environment;

    public function __construct(EnvironmentInterface $environment)
    {
        $this->environment = $environment;
    }

    public function onDocumentParsed(DocumentParsedEvent $event)
    {
        $document = $event->getDocument();
        $walker = $document->walker();
        while ($event = $walker->next()) {
            $node = $event->getNode();

            // Only stop when we first encounter a node
            if (!$event->isEntering()) {
                continue;
            }

            if ($node instanceof Heading && $node->getLevel() === 1) {
                $node->data['attributes']['class'] = 'title-main';
            } elseif ($node instanceof Table) {
                $node->data['attributes']['class'] = 'table';
            }
            // etc.
        }
    }
}
```

The code above is untested but it should get you on the right track.

I do like the idea of creating a more-generic, core piece of functionality that works similar to your example, so I'll keep this open as a feature request for the upcoming 2.0 release :)

armetiz:
awesome!  

I walked throughout the documentation but the event-dispatcher. I didn't expect to find the solution in this section ;p

Thanks !

armetiz:
Note, I'm using `"league/commonmark": "dev-latest"` (commit: 7299ca7).

```
if ($node instanceof Heading) {
[...]

PR #616, opened 2021-05-08
Implement new DefaultAttributesExtension

colinodell:
Implements #605


docs/2.0/extensions/default-attributes.md
```diff
@@ -0,0 +1,121 @@
+---
+layout: default
+title: Default Attributes Extension
```
colinodell, line 3: Needs to be added to the extension list too

PR #617, opened 2021-05-09
Allow unique slugs to be scoped to the document or environment

### LINKED ISSUE #539: Allow heading permalinks to be uniquely scoped per document, environment or manually
This is a follow-up to #531 

_Originally posted by @markcarver in https://github.com/thephpleague/commonmark/issues/531#issuecomment-676478985_:

> Although, the latter could maybe be subverted by maybe creating a new `heading_permalink/unique_slugs ` option that can take multiple values:
>
> `''` - None, uses the normal `SlugNormalizer`
`'document'` - Unique to new documents (resets `UniqueSlugNormalizer ` on `DocumentParsedEvent`)
`'environment'` - Unique to new environments (resets `UniqueSlugNormalizer` on extension construction or `setConfiguration`?)
`'manual'` - `UniqueSlugNormalizer ` remembers everything (i.e. unique to current PHP thread), requires manual reset

_Originally posted by @colinodell in https://github.com/thephpleague/commonmark/issues/531#issuecomment-678830197_:

> I'd be in favor of adding that kind of option, but with only two choices: `'document'` (default) or `'environment'`.
>
> I don't think we should do `''` because it doesn't align with GFM's behavior.
>
> I'm not sure how I feel about `'manual'.  I don't like that it requires global/static storage where two completely different sets of Environments have the ability to impact each other.  I can see why this might be useful - perhaps you have one `Environment` rendering a blog post and another (more-restricted one) to render user comments, and you don't want any collisions in slugs.  But in that case, I think I'd rather see some kind of slug prefix option.
>
> So building off your idea, I'd propose adding two options:
>
> - `heading_permalink/slug_scope`
> - `'document' - Unique to new documents
> - `'environment'` - Unique to new environments
> - `heading_permalink/prefix` - Any arbitrary string which should probably include a `-` character (defaults to `''`)
>
> We'd bring back something like the `UniqueSlugNormalizer` and it would handle:
>
> - Delegating to a child `TextNormalizerInterface` to create the slug
> - Tracking already-used slugs based on the scope in the first option above
> - Appending suffixes to make slugs unique, if needed
> - Maybe some functionality to prepend custom prefixes too?
> - Ability to add, list, and remove a list of "used slugs" (the CRUD idea)
>
> We'd expose an instance of that class via some getter method on the `HeadingPermalinkExtension`.
>
> What do you think?

_Originally posted by @markcarver in https://github.com/thephpleague/commonmark/issues/531#issuecomment-680945935_:

> I'm fine with just `document` and `environment` being the only two values/scenarios as long as the normalizer has some sort of CRUD in place. This would allow a CMS to implement a "manual" implementation easily enough without having to recreate a new environment each time it spits out an email or something with the same content.

colinodell:
This implementation also shares the unique slug generator between extensions to prevent potential conflicts.

Closes #539


joshbruce:
@colinodell: Recommend trying with non-English characters as well. 

We tried ID generation on Marked and ran into a lot of issues with non-English languages, not sure if it's been fixed or whatnot or if CommonMark would run into similar concerns: https://github.com/markedjs/marked

colinodell:
@joshbruce Yeah, we encountered similar issues in the past and updated our existing `SlugNormalizer` to handle those, along with [tests to verify the behavior with other character sets](https://github.com/thephpleague/commonmark/blob/latest/tests/unit/Normalizer/SlugNormalizerTest.php).  The new `UniqueSlugNormalizer` introduced here simply decorates the existing one so I think we're good :relieved:

joshbruce:
Nice! I think the Marked crew decided to get out of the ID generation business and focus on extensibility - if others want to add something, they can. :)

colinodell:
@markhalliwell If you have any availability to review this and offer feedback I'd greatly appreciate it :)

.phpstorm.meta.php
```diff
@@ -66,14 +66,17 @@
         'heading_permalink/insert',
         'heading_permalink/max_heading_level',
         'heading_permalink/min_heading_level',
-        'heading_permalink/slug_normalizer',
+        'heading_permalink/slug_length',
```
markhalliwell, line 69: This should probably be renamed to `slug_normalizer/max_length` as I'm not convinced this is actually `heading_permalink` specific configuration:

https://github.com/thephpleague/commonmark/blob/ae78399dfe9eaecdf64a4a0a1b75112ae0f0c9e3/src/Normalizer/SlugNormalizer.php#L40-L43

It should still probably allow a custom length to be passed in `$context`, but fallback to the configuration value when not provided.

docs/2.0/customization/slug-normalizer.md
```diff
@@ -0,0 +1,92 @@
...
+            public function normalize(string $text, $context = null): string
+            {
+                // TODO: Implement your code here
+            }
+        },
+    ],
+];
+```
+
+This `instance` doesn't need to worry about generating unique slugs, as we automagically decorate your `instance` with `UniqueSlugNormalizer` which provides that functionality.
```
markhalliwell, line 86: I think this automagical feature should be mitigated by a new `slug_normalizer/unique` boolean configuration option that can be opted out of.

While this may be "good enough" for most, if someone is passing their own `slug_normalizer/instance`, it stands to reason they may also need to control what is considered unique.

I don't think we should box anyone in if they need something more than "document" and "environment" scopes.

markhalliwell (commented):
This is a great implementation! A few minor observations that may need to be addressed.

docs/2.0/customization/slug-normalizer.md
markhalliwell: An alternative to a new `slug_normalizer/unique` configuration option, perhaps, would be to check if the provided instance has some sort of `UniqueSlugInterface`.

If it doesn't, then use our own unique wrapper instance? This could allow someone to pass a single instance that both normalizes and disables/implements their own unique slugs.

src/Environment/Environment.php
```diff
@@ -258,6 +265,9 @@ public function addExtension(ExtensionInterface $extension): EnvironmentBuilderI
 
     private function initializeExtensions(): void
     {
+        // Initialize the slug normalizer
+        $this->getSlugNormalizer();
```
markhalliwell, line 269: I know this is needed because it needs to add a `DocumentRenderedEvent` listener, I just don't think we need to retrieve something that may not actually be used.

Perhaps there should be a listener on the `Environment` instead that does something like the following:
```php
    public function onDocumentRendered(DocumentRenderedEvent $event): void
    {
        // Immediately return if there is no slug normalizer to proxy event to.
        if ($this->slugNormalizer === null) {
            return;
        }

        this->slugNormalizer->onDocumentRendered($event);
    }
```

src/Extension/HeadingPermalink/HeadingPermalinkProcessor.php
```diff
@@ -43,36 +44,36 @@ final class HeadingPermalinkProcessor implements ConfigurationAwareInterface
      */
     private $config;
 
-    public function setConfiguration(ConfigurationInterface $configuration): void
+    public function setEnvironment(EnvironmentInterface $environment): void
     {
-        $this->config = $configuration;
+        $this->config         = $environment->getConfiguration();
+        $this->slugNormalizer = $environment->getSlugNormalizer();
```
markhalliwell, line 50: This also instantiates the slug normalizer, regardless if it's needed.

markhalliwell (commented):
A few afterthoughts:

It may be beneficial in the long run to avoid prematurely initializing the slug normalizer instance.

Given that a slug normalizer can be an externally provided instance, it also stands to reason that the instance may be tied to some other resource intensive dependencies.

Perhaps it needs a list of previously used slugs which has to be loaded from a file or DB; perhaps also based on the current user (or another type of entity).

While I understand that this could be checked/loaded when a normalization actually occurs, I also understand that some systems require these types of dependencies to be injected (for performance reasons) when it's a service and the class is constructed.

src/Normalizer/UniqueSlugNormalizer.php
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function normalize(string $text, array $context = []): string
+    {
+        $normalized = $this->innerNormalizer->normalize($text, $context);
+
+        // If it's not unique, add an incremental number to the end until we get a unique version
+        if (\array_key_exists($normalized, $this->alreadyUsed)) {
+            $extension = 0;
```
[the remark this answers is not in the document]
colinodell: Good point! I think I'll rename this to `$suffix`

colinodell:
> It may be beneficial in the long run to avoid prematurely initializing the slug normalizer instance.
>
> Given that a slug normalizer can be an externally provided instance, it also stands to reason that the instance may be tied to some other resource intensive dependencies.
>
> Perhaps it needs a list of previously used slugs which has to be loaded from a file or DB; perhaps also based on the current user (or another type of entity).

I appreciate you sharing that use case.  Respectfully, I don't think that this library should be concerned with how resource-intensive it is to **construct** a normalizer.  We just need the developer to provide us with something that implements `TextNormalizerInterface` - how they construct that object should be left up to them.

For example, a consumer of this library could implement lazy-loading with something like this:

```php
final class LazyLoadedNormalizer implements TextNormalizerInterface
{
    /** @var callable(): TextNormalizerInterface */
    private $constructor;

    /** @var TextNormalizerInterface|null */
    private $normalizer = null;

    /**
     * @param callable(): TextNormalizerInterface $constructor
     */
    public function __construct(callable $constructor)
    {
        $this->constructor = $constructor;
    }

    public function normalize(string $text, array $context = []): string
    {
        if ($this->normalizer === null) {
            $this->normalizer = ($this->constructor)();
        }

        return $this->normalizer->normalize($text, $context);
    }
}
```

And then use service location to ensure the resource-intensive normalizer is only instantiated if/when needed:

```php
$converter = new CommonMarkConverter([
    'slug_normalizer' => [
        'instance' => new LazyLoadedNormalizer(function () {
            // TODO: Construct the normalizer here, or maybe grab it from a DI $container or something
        }),
    ],
]);
```

Another option for people using Symfony (or anything using Symfony's DI container, like Drupal) could be utilizing the [lazy service feature](https://symfony.com/doc/current/service_container/lazy_services.html) to auto-generate a proxy object which the container could then directly inject without using service location.

Therefore, as long as we're careful to not **invoke** any methods on that normalizer, then how the developer chooses to instantiate their objects shouldn't be a concern of this library.

WDYT?

markhalliwell:
Ah! Very true, I didn't think of lazy loading. In that case, yes I agree that the current constructor implementation is fine.

src/Extension/HeadingPermalink/HeadingPermalinkProcessor.php
colinodell: Addressed in my comment below

colinodell:
@markhalliwell I sincerely appreciate all of your feedback and ideas here!  I believe I've addressed everything - let me know if you have any other thoughts or questions.  I know you're a busy man, so I'll wait a week before merging this in case you have feedback, and we can always re-address something later if needed :)  Thanks again!

markhalliwell (approved):
Did another once over and everything LGTM! 👍🏼

PR #619, opened 2021-05-25
Deprecation notice confused me.

Danack:
It just wasn't obvious what the correct way of passing in the config settings was. Could probably be worded better, but English is hard.

### DIFF
--- a/src/GithubFlavoredMarkdownConverter.php
+++ b/src/GithubFlavoredMarkdownConverter.php
@@ -39,7 +39,7 @@ final class GithubFlavoredMarkdownConverter extends MarkdownConverter
                     throw new \RuntimeException('Unable to configure the environment as only ' . Environment::class . ' can be configured after instantiation');
                 }
 
-                @\trigger_error('Configuring custom environments via the constructor is deprecated in league/commonmark v2.0 and will be removed in v3.0; configure it beforehand and create MarkdownConverter with it instead', \E_USER_DEPRECATED);
+                @\trigger_error('Configuring custom environments via the constructor is deprecated in league/commonmark v2.0 and will be removed in v3.0; configure it beforehand and create MarkdownConverter with it instead. And use $environment->mergeConfig($config); if you need to set config settings. ', \E_USER_DEPRECATED);
                 $environment->mergeConfig($config);
             }
 


src/GithubFlavoredMarkdownConverter.php
```diff
@@ -39,7 +39,7 @@ public function __construct(array $config = [], ?EnvironmentInterface $environme
                     throw new \RuntimeException('Unable to configure the environment as only ' . Environment::class . ' can be configured after instantiation');
                 }
 
-                @\trigger_error('Configuring custom environments via the constructor is deprecated in league/commonmark v2.0 and will be removed in v3.0; configure it beforehand and create MarkdownConverter with it instead', \E_USER_DEPRECATED);
+                @\trigger_error('Configuring custom environments via the constructor is deprecated in league/commonmark v2.0 and will be removed in v3.0; configure it beforehand and create MarkdownConverter with it instead. And use $environment->mergeConfig($config); if you need to set config settings. ', \E_USER_DEPRECATED);
```
GrahamCampbell, line 42: `. And` -> `_and`. Sentences generally can't start with "and" (where `_` denotes a space... github doesn't render them for some reason). ;)

GrahamCampbell (changes requested):

src/GithubFlavoredMarkdownConverter.php
colinodell: I've updated the upgrading docs to better show the required changes: https://commonmark.thephpleague.com/2.0/upgrading/consumers/#commonmarkconverter-and-githubflavoredmarkdownconverter-constructors Perhaps we could link to that in the deprecation notice?

(Yes, the old approach is still technically possible in v2.0, but I want to encourage people to switch away from it ASAP)

colinodell: Actually... I just remembered that I decided to deprecate this behavior in 1.6, meaning we can remove it in 2.0.

I therefore think we should do the following:

1. Update the deprecation notice in the `1.6` branch with a link to the upgrading docs
2. Remove the deprecation notice and support for this old approach in the `latest` (essentially 2.0) branch

I'd be happy to implement those changes sometime this week.

colinodell:
Deprecation notice has been updated in commit ceef151a.

PR #627, opened 2021-06-05
Bump minimum PHP version to 7.3

### LINKED ISSUE #625: [2.0] Drop support for PHP 7.2
PHP 7.2 is end-of-life and [only 5% of our users](https://t.co/zKocU3e6Q7) currently use that version.  [PHP 7.3 also has a few features](https://www.zend.com/blog/guide-to-php-73) we might be able to leverage, like case-folding improvements.  Let's therefore drop support for 7.2 in the next major release.

Anyone stuck on 7.1 or 7.2 may continue using 1.x until they are able to upgrade.

colinodell:
Closes #625


GrahamCampbell (approved):
🎉

PR #704, opened 2021-08-09
Add missing import for available classes

lakuapik:
this PR adds import class for `MentionGeneratorInterface`, `Mention` and `AbstractInline` which can be imported from `commonmark` package.


colinodell:
Thank you!

PR #715, opened 2021-09-06
Support of YAML extension for front matter extension

ElGigi:
Add alternative to `symfony/yaml` with YAML extension.

https://www.php.net/manual/book.yaml.php

### DIFF
--- a/docs/2.0/extensions/front-matter.md
+++ b/docs/2.0/extensions/front-matter.md
@@ -19,7 +19,7 @@ composer require league/commonmark
 
 See the [installation](/2.0/installation/) section for more details.
 
-You will also need to install `symfony/yaml` to use this extension:
+You will also need to install `symfony/yaml` or the [YAML extension for PHP](https://www.php.net/manual/book.yaml.php) to use this extension. For `symfony/yaml`:
 
 ```bash
 composer require symfony/yaml
@@ -118,12 +118,16 @@ echo $result->getContent();
 You don't have to parse the entire file (including all the Markdown) if you only want the front matter.  You can either instantiate the front matter parser yourself and call it directly, like this:
 
 ```php
+use League\CommonMark\Extension\FrontMatter\Data\LibYamlFrontMatterParser;
 use League\CommonMark\Extension\FrontMatter\Data\SymfonyYamlFrontMatterParser;
 use League\CommonMark\Extension\FrontMatter\FrontMatterParser;
 
 $markdown = '...'; // TODO: Load some Markdown content somehow
 
+// For `symfony/yaml`
 $frontMatterParser = new FrontMatterParser(new SymfonyYamlFrontMatterParser());
+// For YAML extension
+$frontMatterParser = new FrontMatterParser(new LibYamlFrontMatterParser());
 $result = $frontMatterParser->parse($markdown);
 
 var_dump($result->getFrontMatter()); // The parsed front matter
--- a/src/Extension/FrontMatter/FrontMatterExtension.php
+++ b/src/Extension/FrontMatter/FrontMatterExtension.php
@@ -18,6 +18,7 @@ use League\CommonMark\Event\DocumentPreParsedEvent;
 use League\CommonMark\Event\DocumentRenderedEvent;
 use League\CommonMark\Extension\ExtensionInterface;
 use League\CommonMark\Extension\FrontMatter\Data\FrontMatterDataParserInterface;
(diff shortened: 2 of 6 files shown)


docs/2.0/extensions/front-matter.md
```diff
@@ -19,7 +19,7 @@ composer require league/commonmark
 
 See the [installation](/2.0/installation/) section for more details.
 
-You will also need to install `symfony/yaml` to use this extension:
+You will also need to install `symfony/yaml` or [YAML extension](https://www.php.net/manual/book.yaml.php) to use this extension, for `symfony/yaml`:
```
colinodell, line 22:
```suggestion
You will also need to install `symfony/yaml` or the [YAML extension for PHP](https://www.php.net/manual/book.yaml.php) to use this extension. For `symfony/yaml`:
```

src/Extension/FrontMatter/FrontMatterExtension.php
```diff
@@ -29,7 +30,7 @@ final class FrontMatterExtension implements ExtensionInterface
 
     public function __construct(?FrontMatterDataParserInterface $dataParser = null)
     {
-        $this->frontMatterParser = new FrontMatterParser($dataParser ?? new SymfonyYamlFrontMatterParser());
+        $this->frontMatterParser = new FrontMatterParser($dataParser ?? LibYamlFrontMatterParser::capable() ?? new SymfonyYamlFrontMatterParser());
```
colinodell, line 33: I'm not sure how I feel about this particular change; any users who have `ext-yaml` installed would suddenly start using that parser, even if Symfony's is available and was used by default before.  I suppose that's probably fine though, given that both parsers _should_ largely support the same YAML contents and the extension is faster. So maybe this is okay?  WDYT?

tests/functional/Extension/FrontMatterExtension/FrontMatterExtensionTest.php
```diff
@@ -17,16 +21,30 @@
...
 {
-    private Environment $environment;
+    /**
+     * @return FrontMatterDataParserInterface[]
+     */
+    public function parserProvider(): array
+    {
+        return [
+            [new SymfonyYamlFrontMatterParser()],
+            [new LibYamlFrontMatterParser()],
```
colinodell, line 31: The tests that use `ext-yaml` are failing on PHP 8.1 because `ext-yaml` isn't getting installed.  Perhaps we can somehow mark these tests as skipped?

One possible approach could be providing a second variable that tells the test whether it's possible to use that parser:

```php
public function parserProvider(): array
{
            [new SymfonyYamlFrontMatterParser(), true],
            [new LibYamlFrontMatterParser(), LibYamlFrontMatterParser::isCapable()],
}
```

And in the tests we could have something check that variable:

```php
public function testSomething(FrontMatterDataParserInterface $parser, bool $shouldTest): void
{
    $this->skipIfParserNotAvailable($parser, $shouldTest);

    // ... existing test code goes here
}

// ...

private function skipIfParserNotAvailable(FrontMatterDataParserInterface $parser, bool $shouldTest): void
{
    if (! $shouldTest) {
        $this->markTestSkipped(\sprintf('Cannot test with %s due to missing prerequisites', \get_class($parser)));
    }
```

WDYT?

colinodell (commented):
Overall I really like the idea of providing built-in support for this extension! I just have a few questions/thoughts.

tests/functional/Extension/FrontMatterExtension/FrontMatterExtensionTest.php
ElGigi: Of course, it is better to skip the test if the parser is not available, the modification is pushed.

src/Extension/FrontMatter/FrontMatterExtension.php
ElGigi: For me, it will be faster to parse YAML content of many files with an extension than a library.

But if necessary, one can imagine adding a similar method `SymfonyYamlFrontMatterParser::capable(): SymfonyYamlFrontMatterParser` to provide this parser first.

colinodell: :thinking: I think I'm okay with trying your approach since most users probably won't be affected, and if they are, the only change will likely be faster performance.

colinodell (approved):
LGTM! I'll update the changelog and get this merged over the weekend.  Thanks so much for the contribution! :tada:

PR #729, opened 2021-10-02
Added viewport relative font-size for heading h1.

joelmheim:
Have a look at this suggestion and see if you like it. I'm happy to change it so that it uses fixed sizes for different media queries if you prefer.

### DIFF
--- a/docs/homepage.css
+++ b/docs/homepage.css
@@ -304,7 +304,7 @@ main div > code {
 
 @media screen and (max-width: 549px) {
     .homepage header h1 {
-        font-size:48px;
+        font-size:42px;
     }
 
     .homepage .composer span {
@@ -315,7 +315,7 @@ main div > code {
 @media screen and (min-width: 549px) {
     .homepage header h1 {
         margin:0 auto;
-        font-size:54px;
+        font-size:6.0vw;
     }
 
     .inner-content {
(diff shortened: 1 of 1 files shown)


colinodell:
I love the idea, but unfortunately it causes the text to overflow its container and thus render off-center at larger screen sizes:

image

I think this could be fixed by moving the `<h1>` just outside of its current parent and adding `text-align: center;`:

```diff
 <header>
+    <h1>{{ site.data.project.title }}</h1>
     <div class="inner-content">
-      <h1>{{ site.data.project.title }}</h1>
```

(`docs/_layouts/homepage.html`)

Would you mind giving that a try?

If for any reason that doesn't work we might have to fall back to using fixed sizes instead.  But I really like your `vw` approach so I'm hoping we can keep it :)

joelmheim:
Sure. Can give it a try. Will follow up with new PR.

joelmheim:
Think this new commit might have fixed it. Please have a look. This might be something that can be applied to all the font-sizes to make it resize better as a whole. Let me know what you think.

colinodell (approved):
That looks **really good**, nicely done!  And TYVM for this fix!

> This might be something that can be applied to all the font-sizes to make it resize better as a whole.

If you'd like to open a second PR for that you're welcome to do so, but honestly I think it looks pretty good as-is now :)

PR #742, opened 2021-11-07
CI improvements

colinodell:
Some improvements to the CI pipelines


roave-bc-check.yaml
```diff
@@ -0,0 +1,3 @@
+parameters:
+    ignoreErrors:
+        - '#\[BC\] CHANGED: The number of required arguments for League\\CommonMark\\Extension\\TableOfContents\\TableOfContentsGenerator\#__construct\(\) increased from 4 to 5#'
```
colinodell, line 3: I felt that this violation of SemVer would be acceptable for a few reasons:

- This change was made in 2.0.1 only one week after 2.0.0 was released
- Adoption of 2.0.0 was fairly low at that time
- The overwhelming majority of users (say, 99.999%) would never have a reason to instantiate this class themselves
- It was required to fix a bug in 2.0.0

I am therefore intentionally ignoring this BC break due to the high importance and extremely low impact.

PR #762, opened 2021-12-03
PHP 8.1 deprecated warning

ElGigi:
Prevent deprecated warning on ArrayAccess with PHP 8.1

### DIFF
--- a/src/Util/ArrayCollection.php
+++ b/src/Util/ArrayCollection.php
@@ -93,6 +93,7 @@ final class ArrayCollection implements \IteratorAggregate, \Countable, \ArrayAcc
      *
      * @phpstan-param int $offset
      */
+    #[\ReturnTypeWillChange]
     public function offsetExists($offset): bool
     {
         return \array_key_exists($offset, $this->elements);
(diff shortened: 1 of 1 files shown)


src/Util/ArrayCollection.php
```diff
@@ -120,6 +122,7 @@ public function offsetGet($offset)
      * @phpstan-param int|null $offset
      * @phpstan-param T        $value
      */
+    #[\ReturnTypeWillChange]
```
GrahamCampbell, line 125: This line is not needed. The return type is already `void`.

src/Util/ArrayCollection.php
```diff
@@ -93,6 +93,7 @@ public function count(): int
      *
      * @phpstan-param int $offset
      */
+    #[\ReturnTypeWillChange]
```
GrahamCampbell, line 96: This line is not needed. The return type is already `bool`.

src/Util/ArrayCollection.php
```diff
@@ -136,6 +139,7 @@ public function offsetSet($offset, $value): void
...
```
GrahamCampbell, line 142: This line is not needed. The return type is already `void`.

GrahamCampbell (changes requested):
Only the annotation on `offsetGet` is needed.

ElGigi:
Sorry, changed! A little too fast in my modification.

GrahamCampbell (approved):

colinodell (approved):
Thanks!

PR #767, opened 2021-12-12
Fix typo in mention pattern comment

glensc:


glensc:
Actually, the error itself bothers me:
> throw new InvalidOptionException(\sprintf('Option "mentions/%s/pattern" must not include starting/ending delimiters (like "/")', $name));

when I look at the code, the delimiter is always `"/"`, so the error is incorrect, the only delimiter not allowed is `"/"`, it should be reflected so in error as well.

glensc:
And also, the new logic hardcodes `/i` modifier, but it's not mentioned in the 2.0 migrate guide:
- https://commonmark.thephpleague.com/2.0/upgrading/consumers/#configuration-option-changes

colinodell:
Thanks for the typo fix!

> when I look at the code, the delimiter is always `"/"`, so the error is incorrect, the only delimiter not allowed is `"/"`, it should be reflected so in error as well.

The delimiters are automatically added by the code - the user must therefore not include any type of delimiter (regardless of what character it is) in the partial regex.  I think the bigger issue is that the error can also be thrown in other cases which may be confusing.  I'd be open to rewording this.

> And also, the new logic hardcodes /i modifier, but it's not mentioned in the 2.0 migrate guide

We should probably include a note that patterns will always be case-insensitive. PRs welcome :)

glensc:
> > when I look at the code, the delimiter is always `"/"`, so the error is incorrect, the only delimiter not allowed is `"/"`, it should be reflected so in error as well.
> 
> The delimiters are automatically added by the code - the user must therefore not include any type of delimiter (regardless of what character it is) in the partial regex. I think the bigger issue is that the error can also be thrown in other cases which may be confusing. I'd be open to rewording this.

The problem I'm pointing out, is that it's not delimiters, but it's very specific and only delimiter: `/`

There is no such classifier as "delimiters",
the pcre delimiter can be any character that is 1-byte width, it could even be letter `a`.

edit: Delimiter must not be alphanumeric or backslash, so if you decided to use delimiter as `$`, then how would the user know how to avoid it?

note: backslashes here are for shell not to interpret dollar:

```sh
➔ php -r "var_dump(preg_match('\$b\$', 'b'));"
int(1)
```

my suggestion: say it out clearly, that the delimiter is `/` and it may not be used without escaping.

in other words, I'm allowed to use `;` if I know you use `/` as delimiter!

glensc:
Another idea: probe for delimiters, and fail if not found anything usable (pseudo code):

```py
def find_delimiter(user_pattern):
  for delimiter in ["/", "#", ";", "%"]:
    if not delimiter in user_pattern:
       return delimiter

  raise RuntimeError(f"Could not find any usable delimiter for {user_pattern}")
``

PR #785, opened 2022-01-13
attribute to remove 8.1 deprecation notice

edalzell:
I'm using a CMS that uses Commonmark 1.6, and updating to 2.x is a large endeavor for them at the moment. However this package (1.6) still works great on PHP 8.1, except for one deprecation notice related to the `getIterator` method.

Hoping you'd be willing to merge this into the existing 1.6 branch so we can stop these deprecation errors from flooding our logs.

No functionality added/changed, only an attribute added, which is ignored in PHP < 8.

### DIFF
--- a/src/Util/ArrayCollection.php
+++ b/src/Util/ArrayCollection.php
@@ -72,6 +72,7 @@ class ArrayCollection implements \IteratorAggregate, \Countable, \ArrayAccess
      *
      * @return \ArrayIterator<int|string, mixed>
      */
+    #[\ReturnTypeWillChange]
     public function getIterator()
     {
         return new \ArrayIterator($this->elements);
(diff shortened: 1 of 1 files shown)


edalzell:
I don't understand the failing PHPStan tests, but I didn't do anything around that code so perhaps it's unrelated?

colinodell (approved):
LGTM. Thanks for the contribution!  I'll get this merged, tagged, and released shortly.

I've been planning to end support for v1.6 at the end of this month so your PR is very timely :)

colinodell:
[Version 1.6.7](https://github.com/thephpleague/commonmark/releases/tag/1.6.7) has been released with this fix. Thanks again!

edalzell:
Amazing, thanks for the quick turn around, us Statamic folks (the CMS) really appreciate it...now we can fully support PHP 8.1.

colinodell:
> I don't understand the failing PHPStan tests, but I didn't do anything around that code so perhaps it's unrelated?

Yeah, that was unrelated.  We don't lock to a specific version of PHPStan so sometimes previously-undetected issues crop up when PHPStan adds support for detecting them.  I added a separate commit to fix those just prior to releasing :)

> Amazing, thanks for the quick turn around, us Statamic folks (the CMS) really appreciate it...now we can fully support PHP 8.1.

That's awesome, glad I could help!

PR #794, opened 2022-01-25
Fixed typo in `symfony/deprecation-contracts`

GrahamCampbell:


GrahamCampbell:
I think this was stopping v2 from being resolved (but I may be wrong here. didn't look too closely).

colinodell (approved):
Good catch, I definitely didn't mean for that `v` to be included.  I'll get this merged and released ASAP just in case.  Thanks!

PR #805, opened 2022-02-06
Add Embed extension

### LINKED ISSUE #740: [RFC] Embed Extension
### Description

Provide an extension that generates embed codes for popular services like YouTube, GitHub Gists, etc.

The syntax will look like the image syntax, with the exception that the embed must be on its own line in the Markdown input.

In addition to handling popular services out-of-the-box, users should be able to easily define their own services.

### Syntax

Any `https://` on its own line would be considered embeddable content:

```md
https://youtube.com/watch?v=zbA62fvGGyw

https://youtu.be/zbA62fvGGyw

https://gist.github.com/colinodell/ffdd892cedc43d07597cb73a9a72e668
```

But this would not, since the URL isn't on its own line:

```md
Check out my favorite video: https://youtube.com/watch?v=zbA62fvGGyw
```

colinodell:
Implements the ability to easily embed rich content from other websites via OEmbed and other such APIs.

## Syntax

The syntax for using this extension is straight-forward - just drop a link on its own line anywhere in the Markdown document:

```md
Check out this video!

https://www.youtube.com/watch?v=dQw4w9WgXcQ
```

More details about the syntax and usage of the extension can be found in the [proposed documentation for this extension](https://github.com/thephpleague/commonmark/blob/embed-extension/docs/2.3/extensions/embed.md).

## Previously Proposed Syntax

I initially proposed reusing the image syntax for this feature:

```md
![](https://www.youtube.com/watch?v=dQw4w9WgXcQ)
```

But on further reflection this had some significant downsides:

### Performance

It would be impossible to differentiate between an image and an embeddable resource without making an HTTP request each time image-like syntax appeared.  Even if we managed to make those calls in parallel, it still requires extra wasted overhead.

### Violates the spirit of Markdown

As John Gruber said:

> The overriding design goal for Markdown’s formatting syntax is to make it as readable as possible. The idea is that a Markdown-formatted document should be publishable as-is, as plain text, without looking like it’s been marked up with tags or formatting instructions.

### Violates the spirit of OEmbed

The whole point of OEmbed is that users simply drop a link somewhere:

> oEmbed is a format for allowing an embedded representation of a URL on third party sites. The simple API allows a website to display embedded content (such as photos or videos) when a user posts a link to that resource

The experience should be seamless, just like it is when posting URLs on Twitter, Slack, etc.

## Feedback Wanted

Any/all feedback on this feature would be greatly appreciated!

(Closes #740)


ArabCoders:
I honestly think directly parsing the URLs might be a really bad idea. sometimes you just want put a URL in the text.

I think the old proposal was on correct track as this is not a standard, i think you may even go further by using specific tags. like

:video this is video example :image this is image example as such you dont have to parse the whole body and user get to choose if they want make it embeddable or something else, this extension could as well put the two functionality and let the user choose. and the domains should be extendable, for self hosted etc.

# Violates the spirit of OEmbed

This might be applicable if you have some sort of URL textbox specific for URLs. However markdown usually used in text area with many many different text some which might look like a URL. In this case i dont think it's applicable.

Thank you.

kleinmann:
I like limiting the embeds to standalone links and ignoring them in any other elements like inline with text.

I'm not sure if users would simultaneously want to embed stuff, but also show text links as separate blocks in the same document. I'd probably put the link in something like a blockquote, if I want to emphasize the link itself in such cases.
Those would then be excluded from the proposed embed solution, right?

I also follow the argument of not using any additional syntax (like image embeds) for this. This is comparable to the autolink extension for me, in that deactivating the extension still renders a perfectly fine document without extension-specific artifacts.

Looking forward to this extension 🙂

colinodell:
> I honestly think directly parsing the URLs might be a really bad idea. sometimes you just want put a URL in the text.

This Embed extension will only attempt to embed URLs that:

1. Are the only text within a paragraph (must have an empty line before and after, and no other text or syntax on the same line)
2. Start with `https://`
3. Don't use the `[label](url)` or `<url>` syntax
 
Normally these URLs would be rendered as plain text (unless you use the [Autolink extension](https://commonmark.thephpleague.com/2.2/extensions/autolinks/)).

If the given URL does not contain embeddable content, it will be rendered as a normal link instead (just like how Autolink would handle it).

And lastly, users would have to opt-in to using this extension - it would not be enabled by default.  If somebody does enable this and they still want to place a URL on its own line without embedding it, they could use the `<url>` syntax instead to make it a link.

some0ne2:
Will there be a way to "disable" this behavior / keep the old style?

All the best;

colinodell:
> Will there be a way to "disable" this behavior / keep the old style?

Yes, this is an optional extension which will not be enabled by default.  Only users who explicitly add and configure this additional extension will get this behavior.

PR #808, opened 2022-02-08
Image labels as attributes; avoid double-escaping

### LINKED ISSUE #806: Image alt attributes are double-escaped
### Version(s) affected

2.1.1

### Description

This also affects 1.6.7.

Input:
```md
![The Mass Gravel & Sand area.](/assets/guides/107/fo4_mass_gravel_and_sand_header.jpg)
```

Output:
```html
<p><img src="/assets/guides/107/fo4_mass_gravel_and_sand_header.jpg" alt="The Mass Gravel &amp;amp; Sand area." /></p>
```

The left and right angle brackets are also double-escaped.

### How to reproduce

Create an image with alt text that contains a character that would need to be escaped.

### Possible solution

_No response_

### Additional context

_No response_

### Did this project help you today? Did it make you happy in any way?

<3

GarrettAlbright:
Fixes #806.

Image labels (alt text) were represented as child nodes on the Image node which are rendered, then the text stuffed into an attribute on an HtmlElement. This was resulting in special characters (particularly quotes, angle brackets, and ampersands) in the label to be double-escaped.

While it somewhat makes sense for labels to be children of Image since, according to spec, they are themselves inline elements, aside from the double-escaping issue, this is also a break from how other attributes are handled on nodes and had some hacky code to deal with it. This replacement code has its own weird quirks (see CloseBracketParser.php's diff) but achieves its goals.

A test case was altered to add code for an image with special characters in its labels and other cases were adjusted to deal with labels no longer being children of images.

Note that the message on commit eb69572 is incorrect; we're changing behavior of image labels, not image titles. (Hopefully I can be forgiven for mixing the two up frequently.)

If you like video games, it may be of interest that [Gamer Guides](https://gamerguides.com) sponsored my development of this fix.

### DIFF
--- a/src/Extension/CommonMark/Node/Inline/Image.php
+++ b/src/Extension/CommonMark/Node/Inline/Image.php
@@ -20,17 +20,25 @@ use League\CommonMark\Node\Inline\Text;
 
 class Image extends AbstractWebResource
 {
+    protected string $label;
     protected ?string $title = null;
 
-    public function __construct(string $url, ?string $label = null, ?string $title = null)
+    public function __construct(string $url, ?string $label = '', ?string $title = null)
     {
         parent::__construct($url);
 
-        if ($label !== null && $label !== '') {
-            $this->appendChild(new Text($label));
-        }
-
+        $this->label = $label ?? '';
         $this->title = $title;
+    }
+
+    public function getLabel(): string
+    {
+        return $this->label;
+    }
+
+    public function setLabel(?string $label): void
+    {
+        $this->label = $label;
     }
 
     public function getTitle(): ?string
--- a/src/Extension/CommonMark/Node/Inline/Image.php
+++ b/src/Extension/CommonMark/Node/Inline/Image.php
@@ -16,8 +16,6 @@ declare(strict_types=1);
 
 namespace League\CommonMark\Extension\CommonMark\Node\Inline;
 
-use League\CommonMark\Node\Inline\Text;
-
 class Image extends AbstractWebResource
 {
     protected string $label;
--- a/src/Extension/CommonMark/Parser/Inline/CloseBracketParser.php
+++ b/src/Extension/CommonMark/Parser/Inline/CloseBracketParser.php
@@ -105,8 +105,7 @@ final class CloseBracketParser implements InlineParserInterface, EnvironmentAwar
                     $inline->setLabel($inline->getLabel() . $label->getTitle());
                 } elseif ($label instanceof Image) {
                     $inline->setLabel($inline->getLabel() . $label->getLabel());
-                }
-                else {
+                } else {
                     $literal = $label->getLiteral();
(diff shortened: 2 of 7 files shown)


src/Extension/CommonMark/Node/Inline/Image.php
```diff
@@ -16,23 +16,29 @@
...
 
-use League\CommonMark\Node\Inline\Text;
-
 class Image extends AbstractWebResource
 {
+    protected string $label;
     protected ?string $title = null;
 
-    public function __construct(string $url, ?string $label = null, ?string $title = null)
+    public function __construct(string $url, ?string $label = '', ?string $title = null)
```
GrahamCampbell, line 24: I don't think this change is correct.

GrahamCampbell (changes requested):

colinodell:
Thank you so much for working on this!

Reviewing the spec closely, I came across this:

> Syntax for images is like the syntax for links, with one difference. Instead of link text, we have an image description. The rules for this are the same as for link text [...] An image description has inline elements as its contents. When an image is rendered to HTML, this is standardly used as the image’s alt attribute.
>
> Though this spec is concerned with parsing, not rendering, it is recommended that in rendering to HTML, only the plain string content of the image description be used.

And the definition of link text:

> A link text consists of a sequence of zero or more inline elements

Given this, I'd strongly prefer that we keep the description's inline elements as-is during parsing and address this only within the render step. That approach would keep us closely aligned with the spec and ensure that our AST will continue to look virtually identical to the reference parsers.

I think the problem with my original implementation was attempting to render ALL of the children (and then use kludgy regular expressions to sanitize it). Instead of rendering all children, we could have the image renderer iterate through the children and only include things like `Text` node literals, ignoring other node types we don't want. This is pretty much how the Java library does it - see https://github.com/commonmark/commonmark-java/blob/bb5df203d6e0b84e7537d14043d8837011b18ae0/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java#L183-L185 and https://github.com/commonmark/commonmark-java/blob/bb5df203d6e0b84e7537d14043d8837011b18ae0/commonmark/src/main/java/org/commonmark/renderer/html/CoreHtmlNodeRenderer.java#L297-L319.

Would you be open to giving that approach a try?

src/Extension/CommonMark/Node/Inline/Image.php
GarrettAlbright: My logic with this change is that, as far as I can tell, it's impossible for an  tag created with MD/CM to not have an alt attribute. `` results in ``.

GarrettAlbright:
@colinodell 

Okay, I've got something working using the approach you suggested. Since it's a pretty different direction than this branch/PR, do you think it would be better to make a new PR, or should I just go ahead and merge it into this brach?

GarrettAlbright:
Closing for #810 then.

PR #810, opened 2022-02-09
Change method to get label (alt text) of images; avoid double-escaping

GarrettAlbright:
Fixes https://github.com/thephpleague/commonmark/issues/806.

This adds a getLabel() method to the Image node which attempts to get a text representation of the child nodes, and calls it to generate the "alt" attribute for the final `` tag. This changes the previous approach which involved rendering the children and then extracting the text via regular expressions from the generated HTML. This was also causing special characters in the alt text to be double-escaped.

If you like video games, it may be of interest that [Gamer Guides](https://gamerguides.com/) sponsored my development of this fix.

### DIFF
--- a/src/Extension/CommonMark/Node/Inline/Image.php
+++ b/src/Extension/CommonMark/Node/Inline/Image.php
@@ -16,7 +16,10 @@ declare(strict_types=1);
 
 namespace League\CommonMark\Extension\CommonMark\Node\Inline;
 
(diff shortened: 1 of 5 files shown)


colinodell:
Thanks for the fix!

I did a little refactoring to move all the logic into the renderer, but otherwise, this looks great to me.  I'll fix the unrelated Psalm errors and have this released shortly :)

PR #845, opened 2022-03-30
Update .gitattributes

PaolaRuby:
Remove unnecessary files from vendor


colinodell (approved):
Thanks for the change! I've removed `CHANGELOG.md` from this list, as I do want the latest changelog to be included, but everything else looks great :) Thanks again!

PR #853, opened 2022-04-07
Fix position of description of tryContinue

EreMaijala:
Moved the description of `tryContinue()` from under the heading for `closeBlock()`.


colinodell (approved):

Issue #884, opened 2022-06-06
Setting allowed_domains for the embed extension causes an array key error


mobilisedonline:
### Version(s) affected

2.3.2

### Description

If I only allow twitter.com and youtube.com in the allowed_domains config but a github.com URL appears first in the markdown then the array keys will not match in the updateEmbeds method of OscaroteroEmbedAdapter causing an array key error.

I believe the problem is caused by the array_filter function in DomainFilteringAdapter leaving a gap in the embeds array when applying the allowed_domains list.

### How to reproduce

Set the embed config to:

```
'embed' => [
    'adapter' => new OscaroteroEmbedAdapter($embedLibrary),
    'allowed_domains' => ['youtube.com', 'twitter.com'],
    'fallback' => 'link',
],
```
Set the markdown content to:

> https://github.com/laravel
> 
> https://twitter.com/daringfireball/status/1532083567156572160
> 
> https://www.youtube.com/watch?v=v1iZtBM23bY

### Possible solution

Either reindex the embeds array after applying the allowed_domains filter or loop over the embeds array within the extractors loop to match the correct embed code rather than relying on the array indexes matching.

PR #885, opened 2022-06-06
Reindex  array to remove gaps after applying allowed_domains list

mobilisedonline:
Fix for #884

### DIFF
--- a/src/Extension/Embed/DomainFilteringAdapter.php
+++ b/src/Extension/Embed/DomainFilteringAdapter.php
@@ -33,9 +33,9 @@ class DomainFilteringAdapter implements EmbedAdapterInterface
      */
     public function updateEmbeds(array $embeds): void
     {
-        $this->decorated->updateEmbeds(\array_filter($embeds, function (Embed $embed): bool {
+        $this->decorated->updateEmbeds(\array_values(\array_filter($embeds, function (Embed $embed): bool {
             return \preg_match($this->regex, $embed->getUrl()) === 1;
-        }));
+        })));
     }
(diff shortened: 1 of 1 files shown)


colinodell (approved):
Thanks for the fix!

Would you mind also adding a regression test (either to [DomainFilteringAdapterTest.php](https://github.com/thephpleague/commonmark/blob/2.3/tests/unit/Extension/Embed/DomainFilteringAdapterTest.php) or perhaps an integration test with both adapters) that confirms the fix works as expected?  If you're not sure how or don't have the bandwidth I'll gladly add it myself - it might just take a few days for me to find the time.

mobilisedonline:
> Thanks for the fix!
> 
> Would you mind also adding a regression test (either to [DomainFilteringAdapterTest.php](https://github.com/thephpleague/commonmark/blob/2.3/tests/unit/Extension/Embed/DomainFilteringAdapterTest.php) or perhaps an integration test with both adapters) that confirms the fix works as expected? If you're not sure how or don't have the bandwidth I'll gladly add it myself - it might just take a few days for me to find the time.

I could see from testing the fix on the site I'm developing that the gaps in the array were removed but I haven't been able to get my head around how to write a test to prove it.

colinodell:
No problem! I truly don't mind adding this myself :-)

colinodell:
Test added via effe03b.  I modified it to use a mock so we could more easily verify the actual array being passed to the decorated instance.

Thanks again for catching and resolving this!  I'll release v2.3.3 with this fix momentarily.

PR #893, opened 2022-07-01
Adds missing front matter extension to the documentation menu

froschdesign:
The extension "Front Matter" is listed in the ["Extensions Overview"](https://commonmark.thephpleague.com/2.3/extensions/overview/) but is missing in the documentation menu for all versions since 2.0.0.


colinodell (approved):
Great catch, and thanks for the fix!

PR #924, opened 2022-09-20
[WIP]: move trim from parser to renderer

MrSpoocy:
There is a PHPCS bug that I don't understand what is wrong with it: "Incorrect annotations group."

Locally there are still 4 PHPUnit tests which fail. But I am not sure if there is not a fundamental error in thinking there. 
Because in XML there is the flag "xml:space="preserve"". It could even be that the test needs to be rethought.

This pull-request is just to give you the opportunity to think about my suggestion (https://github.com/thephpleague/commonmark/discussions/923) and see how I implemented it.


colinodell:
Closing this based on our discussion in #923

PR #931, opened 2022-09-27
Added MarkerExtension to Community Extensions List

noah1400:
It allows users to use == in order to indicate text that should be rendered within `mark` tags.


colinodell (approved):
This is great! Thanks for creating the extension and listing it here :)

PR #939, opened 2022-10-21
Allow headingpermlink attaching to the heading

MrSpoocy:
Additionally:
- [x] Always add tests and ensure they pass.
- [x] Avoid breaking backward compatibility
- [x] New features and deprecations must be submitted against the "main" branch
- [x] Absolutely nothing changes in the default configuration.

https://commonmark.thephpleague.com/2.3/extensions/heading-permalinks/
`This extension makes all of your heading elements (<h1>, <h2>, etc) linkable so that users can quickly grab a link to that specific part of the document - almost like the headings in this documentation!`

Yup and this update, allows setting the anchor directly on the headline. Just like you have in your documentation xD

I also think that maybe there should be the possibility to disable the <a> tag completely but then have the anchor in the headline.

I put a lot of effort to consider everything, UnitTest, documentation updated and no breaking changes. I hope that the pull request gets your approval.

### DIFF
--- a/docs/2.3/extensions/heading-permalinks.md
+++ b/docs/2.3/extensions/heading-permalinks.md
@@ -38,6 +38,8 @@ $config = [
     'heading_permalink' => [
         'html_class' => 'heading-permalink',
         'id_prefix' => 'content',
+        'attach_heading' => false,
+        'heading_class' => 'heading-anchor'
         'fragment_prefix' => 'content',
         'insert' => 'before',
         'min_heading_level' => 1,
@@ -71,6 +73,14 @@ The value of this nested configuration option should be a `string` that you want
 ### `id_prefix`
 
 This should be a `string` you want prepended to HTML IDs.  This prevents generating HTML ID attributes which might conflict with others in your stylesheet.  A dash separator (`-`) will be added between the prefix and the ID.  You can instead set this to an empty string (`''`) if you don't want a prefix.
+
+### `attach_heading` (since 2.3.6)
+
+If this value is `true`, the `id` attributes will be written to the <h> tag instead of the <a>.
+
+### `heading_class` (since 2.3.6)
+
+If attach_heading is `true`, this class is added to the <h> tag.
 
 ### `fragment_prefix`
 
--- a/docs/2.3/extensions/heading-permalinks.md
+++ b/docs/2.3/extensions/heading-permalinks.md
@@ -38,8 +38,6 @@ $config = [
     'heading_permalink' => [
         'html_class' => 'heading-permalink',
         'id_prefix' => 'content',
-        'attach_heading' => false,
-        'heading_class' => 'heading-anchor'
         'fragment_prefix' => 'content',
         'insert' => 'before',
         'min_heading_level' => 1,
@@ -73,14 +71,6 @@ The value of this nested configuration option should be a `string` that you want
 ### `id_prefix`
 
 This should be a `string` you want prepended to HTML IDs.  This prevents generating HTML ID attributes which might conflict with others in your stylesheet.  A dash separator (`-`) will be added between the prefix and the ID.  You can instead set this to an empty string (`''`) if you don't want a prefix.
-
-### `attach_heading` (since 2.3.6)
-
-If this value is `true`, the `id` attributes will be written to the <h> tag instead of the <a>.
-
-### `heading_class` (since 2.3.6)
-
-If attach_heading is `true`, this class is added to the <h> tag.
 
 ### `fragment_prefix`
 
--- a/src/Extension/HeadingPermalink/HeadingPermalinkExtension.php
+++ b/src/Extension/HeadingPermalink/HeadingPermalinkExtension.php
@@ -31,6 +31,8 @@ final class HeadingPermalinkExtension implements ConfigurableExtensionInterface
             'max_heading_level' => Expect::int()->min(1)->max(6)->default(6),
             'insert' => Expect::anyOf(HeadingPermalinkProcessor::INSERT_BEFORE, HeadingPermalinkProcessor::INSERT_AFTER)->default(HeadingPermalinkProcessor::INSERT_BEFORE),
             'id_prefix' => Expect::string()->default('content'),
+            'attach_heading' => Expect::bool()->default(false),
+            'heading_class' => Expect::string()->default('heading-anchor'),
             'fragment_prefix' => Expect::string()->default('content'),
             'html_class' => Expect::string()->default('heading-permalink'),
             'title' => Expect::string()->default('Permalink'),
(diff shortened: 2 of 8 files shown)


MrSpoocy:
And now I had to realize that I had created the branch from 2.3 instead of the main :(

MrSpoocy:
`[BC] CHANGED: The number of required arguments for League\CommonMark\Extension\HeadingPermalink\HeadingPermalink#__construct() increased from 1 to 3
`

Is it a breaking change if the class was final (can not extend) and the first parameter is the same?

colinodell:
> Yup and this update, allows setting the anchor directly on the headline.

Could you explain the value here?  Like, I fully understand *what* this feature allows, but I'm having trouble understanding *why* this would be important to allow.  What's the use case here?  Knowing that will help me to give this a proper review :)

> Is it a breaking change if the class was final (can not extend) and the first parameter is the same?

Yes, because any external code that used to instantiate a new `HeadingPermalink()` with one parameter will no longer work.

MrSpoocy:
It first ensures the separation of the anchor and the link to it. Currently, for example, the icon for an anchor is set by configuration and also decided where it is set.  But this decision should be made by the desginer e.g. via CSS.

Further I would like to add the possibility that the <a> tag is not generated, that allows users to link within the own document to the headline (exactly that is my goal without writing extra a new extension).

It sounds more plausible to me to use the already existing extension for Headline instead of parsing the document again and adding more HTML.

docs/2.3/extensions/heading-permalinks.md
```diff
@@ -72,6 +74,14 @@ The value of this nested configuration option should be a `string` that you want
...
 This should be a `string` you want prepended to HTML IDs.  This prevents generating HTML ID attributes which might conflict with others in your stylesheet.  A dash separator (`-`) will be added between the prefix and the ID.  You can instead set this to an empty string (`''`) if you don't want a prefix.
 
+### `attach_heading` (since 2.3.6)
+
+If this value is `true`, the `id` attributes will be written to the <h> tag instead of the <a>.
+
+### `heading_class` (since 2.3.6)
+
+If attach_heading is `true`, this class is added to the <h> tag.
+
```
colinodell, line 84: This will have to be released in 2.4.0 since it's technically a new feature, so let's move all the documentation changes to `docs/2.4/`

src/Extension/HeadingPermalink/HeadingPermalinkExtension.php
```diff
@@ -31,6 +31,8 @@ public function configureSchema(ConfigurationBuilderInterface $builder): void
             'max_heading_level' => Expect::int()->min(1)->max(6)->default(6),
             'insert' => Expect::anyOf(HeadingPermalinkProcessor::INSERT_BEFORE, HeadingPermalinkProcessor::INSERT_AFTER)->default(HeadingPermalinkProcessor::INSERT_BEFORE),
             'id_prefix' => Expect::string()->default('content'),
+            'attach_heading' => Expect::bool()->default(false),
+            'heading_class' => Expect::string()->default('heading-anchor'),
```
colinodell, line 35: I'd strongly prefer that this defaults to an empty string since there's a small possibility that users might have existing CSS styles targeting `.heading-anchor` for other purposes and so upgrading `league/commonmark` to 2.4 may unexpectedly break site styles.

src/Extension/HeadingPermalink/HeadingPermalinkProcessor.php
```diff
@@ -45,27 +45,40 @@ public function setEnvironment(EnvironmentInterface $environment): void
...
         $slug = $this->slugNormalizer->normalize($text, [
             'node' => $heading,
             'length' => $slugLength,
         ]);
 
-        $headingLinkAnchor = new HeadingPermalink($slug);
+        if ($attachHeading) {
+            $heading->data->set('attributes/id', $idPrefix . $slug);
+            if ($headingClass !== '') {
+                $heading->data->append('attributes/class', $this->config->get('heading_permalink/heading_class'));
```
colinodell, line 77: I think we should allow the heading class to always be configurable - even when they aren't using the new `attach_heading` config option.  (We can use the documentation to explain that the `heading_class` option is especially helpful when the `attach_heading` behavior is enabled, but that it can still be used more generally if desired)

colinodell (changes requested):
Got it, thank you for those extra details!  I agree, I think changes like this could be useful :)

Would you mind making a few changes to the proposed implementation?

src/Extension/HeadingPermalink/HeadingPermalinkExtension.php
```diff
@@ -31,6 +31,8 @@ public function configureSchema(ConfigurationBuilderInterface $builder): void
             'max_heading_level' => Expect::int()->min(1)->max(6)->default(6),
             'insert' => Expect::anyOf(HeadingPermalinkProcessor::INSERT_BEFORE, HeadingPermalinkProcessor::INSERT_AFTER)->default(HeadingPermalinkProcessor::INSERT_BEFORE),
             'id_prefix' => Expect::string()->default('content'),
+            'attach_heading' => Expect::bool()->default(false),
```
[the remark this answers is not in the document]
MrSpoocy: I prefer then `apply_id_to_heading`, because the possibility to allow a string (even if it is a constant) would imply that there is a choice, but so far, there is only yes or no.

src/Extension/HeadingPermalink/HeadingPermalinkExtension.php
MrSpoocy: Up to you, I will empty it

src/Extension/HeadingPermalink/HeadingPermalinkProcessor.php
MrSpoocy: I am always for more options and flexibility :)

MrSpoocy:
If I'm already working on it, should I include the ability to disable the `<a>` tag already? So that the user can always decide for himself where in the document there is a link on the anchor everywhere.

If so, extend the `insert` option with a constant `HeadingPermalinkProcessor::INSERT_NONE`?

colinodell:
> If so, extend the insert option with a constant HeadingPermalinkProcessor::INSERT_NONE?

Yeah, that makes sense to me! Let's do it :)

src/Extension/HeadingPermalink/HeadingPermalinkProcessor.php
```diff
@@ -45,27 +45,40 @@ public function setEnvironment(EnvironmentInterface $environment): void
...
+            }
+        }
+
+        $headingLinkAnchor = new HeadingPermalink($slug, $attachHeading);
```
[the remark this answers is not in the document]
MrSpoocy: We can't move everything to the renderer, unfortunately, because if we use "INSERT_NONE", I would have just omitted the `$heading->appendChild($headingLinkAnchor);` now. 

This way the renderer will never process.

colinodell (approved):

colinodell:
Thanks for the contribution! This will be released in 2.4.0.

Issue #943, opened 2022-10-25
Allow simplified input for attribute values


MrSpoocy:
### Description

Related to the https://commonmark.thephpleague.com/2.3/extensions/attributes/ extension, I would like to make an adjustment. In order to make it easier for the users, I would be a bit more generous when entering values for atttributes (can be activated via configuration).

For all key-value values, where the value only consists of the following characters `A-Za-z0-9-_.` the user does not need to write quotation marks (they are set in the HTML anyway).

It is only an improvement for the operability. What is your opinion?
I would make this adjustment as part of the change in #942.

### Example

`[Goolge can help you](https://www.google.de){target=_blank}` -> `<a href="https://www.google.de" target="_blank">Goolge can help you</a>`

`### Headline {class=big}` -> `<h3 class="big">Headline</h3>`

`Total crazy{class=cms.highlight}` -> `<p class="cms.highlight">Total crazy</p>`

colinodell:
Looking at [one of our test cases](https://raw.githubusercontent.com/thephpleague/commonmark/2.3/tests/functional/Extension/Attributes/data/table_attributes.md) it seems like this might already be supported:

https://github.com/thephpleague/commonmark/blob/32891d1aa3a7ad4cad6a60177d3b4bb8dbdff96b/tests/functional/Extension/Attributes/data/table_attributes.md?plain=1#L1-L4

  Would you be open to adding [some more test cases](https://github.com/thephpleague/commonmark/blob/32891d1aa3a7ad4cad6a60177d3b4bb8dbdff96b/tests/unit/Extension/Attributes/Util/AttributesHelperTest.php#L40) to confirm whether this works as expected?  We can then commit those new test cases if they pass or explore options (like your proposal) if they don't currently work.

MrSpoocy:
Then maybe it's a bug I found. I will extend the tests tomorrow morning. What I can already say, with me the following did not go (the point at the end provides for problems):

```
This is
a test [Google](https://www.google.de){target=_blank}.

Second line
```

PR #944, opened 2022-10-26
Add some more attributes test

MrSpoocy:
Related to #943, here's a few more test. One of them also fails at the moment because there seems to be a bug.


src/Util/RegexHelper.php
```diff
@@ -43,7 +43,7 @@ final class RegexHelper
     public const PARTIAL_TAGNAME               = '[a-z][a-z0-9-]*';
     public const PARTIAL_BLOCKTAGNAME          = '(?:address|article|aside|base|basefont|blockquote|body|caption|center|col|colgroup|dd|details|dialog|dir|div|dl|dt|fieldset|figcaption|figure|footer|form|frame|frameset|h1|head|header|hr|html|iframe|legend|li|link|main|menu|menuitem|nav|noframes|ol|optgroup|option|p|param|section|source|summary|table|tbody|td|tfoot|th|thead|title|tr|track|ul)';
     public const PARTIAL_ATTRIBUTENAME         = '[a-z_:][a-z0-9:._-]*';
-    public const PARTIAL_UNQUOTEDVALUE         = '[^"\'=<>`\x00-\x20]+';
+    public const PARTIAL_UNQUOTEDVALUE         = '[^"\'=<>}`\x00-\x20]+';
```
colinodell, line 46: This constant is also used to parse HTML and according to the spec a `}` is allowed in unquoted values:

> Attributes are placed inside the start tag, and consist of a name and a value, separated by an "=" character. The attribute value can remain unquoted if it doesn't contain ASCII whitespace or any of " ' ` = < or >. Otherwise, it has to be quoted using either single or double quotes. The value, along with the "=" character, can be omitted altogether if the value is the empty string.

So we'll need to find another way to fix the issue you found.

colinodell:
Thanks for adding these tests! They were super helpful in pinpointing the cause of this issue.  I'm going to merge an alternate solution (that includes your new tests) in #945.

PR #952, opened 2022-12-10
Fix mbstring calls without explicit encoding

### LINKED ISSUE #951: Invalid multibyte support
### Version(s) affected

2.3.7

### Description

If `mb_internal_encoding()` is anything but UTF-8 and a multibyte character is used before an inline link, output will be garbage such as `ABC ÀÀ [label](url)` -> `ABC ÀÀ  [l[bel](url)`.

Bug is caused by a `mb_strlen` call not passing encoding in `InlineParserEngine`.

A quick regex search found 11 similar calls. Some may be safe because they don't expect multibyte input (URI cannot be multibyte in markdown). In this case they should be converted to non multibyte calls to avoid the confusion.

### How to reproduce

```
mb_internal_encoding('iso-8859-1');
(new CommonMarkConverter)->convert('AAA ÀÀ [label][(https://url)');
```

colinodell:
Fixes parsing issues when `mb_internal_encoding()` is set to something other than `UTF-8`

Fixes #951

### DIFF
--- a/.github/workflows/backwards-compatibility.yml
+++ b/.github/workflows/backwards-compatibility.yml
@@ -18,6 +18,13 @@ jobs:
               with:
                   fetch-depth: 0
 
+            # User in the container seems to be different than the cloned repo's owner.
+            # Git doesn't like that as the repo will then be unusable by the owner.
+            # We don't care about this here since this is only  used for running one test.
+            # See https://github.com/actions/runner/issues/2033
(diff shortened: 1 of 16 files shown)


colinodell:
CI is failing due to https://github.com/Nyholm/roave-bc-check-docker/issues/32 but it should be safe to merge anyway

PR #956, opened 2023-01-26
Fix autolink extension URL detection

LeSuisse:
The extension was not detecting URI with a `_` symbol correctly.

The issue was already fixed in [the regex used by the Symfony project](https://github.com/symfony/symfony/blob/7b0ca3ab5678232f876a95991b66778f4aaa33e3/src/Symfony/Component/Validator/Constraints/UrlValidator.php#L25-L41).
The regex used by the extension is updated to integrate all the changes that been made to it.


colinodell (approved):
LGTM, thank you for the fix!

Issue #958, opened 2023-02-15
Table centering is producing `align="center"` instead of `style="text-align: center"`


ptmkenny:
### Version(s) affected

2.3.8

### Description

According to the [documentation on the Table extension](https://commonmark.thephpleague.com/2.3/extensions/tables/):

```
---|:----------:|----------:
td | td         | td
```

should produce:

```
<table>
<thead>
<tr>
<th style="text-align: left">th</th>
<th style="text-align: center">th(center)</th>
<th style="text-align: right">th(right)/th>
</tr>
</thead>
```

However, with version 2.3.8, I see the following:

```
<table>
<thead>
<tr><th>th</th>
<th align="center">th(center)</th>
<th align="right">th(right)</th>
</tr>
</thead>
```

### How to reproduce

Input:

```
---|:----------:|----------:
td | td         | td
```

Config:

Running this on Drupal 10 without using the Drupal module (I am calling commonmark directly in my code).

```
/**
 * Converts CommonMark-compatible Markdown to HTML.
 */
final class MyModuleCommonMarkConverter extends MarkdownConverter {

  /**
   * Create a new Markdown converter pre-configured for CommonMark.
   */
  public function __construct() {
    $environment = new Environment($this->getConfig());
    $environment->addExtension(new CommonMarkCoreExtension());
    $environment->addExtension(new DefaultAttributesExtension());
    $environment->addExtension(new ExternalLinkExtension());
    $environment->addExtension(new HeadingPermalinkExtension());
    $environment->addExtension(new StrikethroughExtension());
    $environment->addExtension(new TableExtension());

    parent::__construct($environment);
  }

  /**
   * Get the environment.
   */
  public function getEnvironment(): Environment {
    \assert($this->environment instanceof Environment);

    return $this->environment;
  }

  /**
   * Generate config for CommonMark.
   *
   * This must be a function because using variables in defaultAttributes
   * will error: 'Expression is not allowed as field default value'.
   *
   * https://commonmark.thephpleague.com/2.3/configuration/.
   *
   * @return array
   *   The CommonMark config.
   */
  private function getConfig(): array {
    return [
      'default_attributes' => [
        Heading::class => [
          'class' => static function (Heading $node) {
            if ($node->getLevel() === 1) {
              return 'title';
            }
            else {
              return NULL;
            }
          },
        ],
        /* Link::class => [
             'class' => 'btn btn-link',
             'target' => '_blank',
           ], */
      ],
      'renderer' => [
[...]

PR #959, opened 2023-02-15
Table alignment config option

colinodell:
This PR introduces a new config option to control whether table cell alignment is rendered using `style` or `align` attributes. See #958 for more context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
     - `heading_permalink/apply_id_to_heading` - When `true`, the `id` attribute will be applied to the heading element itself instead of the `<a>` tag
     - `heading_permalink/heading_class` - class to apply to the heading element
     - `heading_permalink/insert` - now accepts `none` to prevent the creation of the `<a>` link
+- Added new `table/alignment_attributes` configuration option to control how table cell alignment is rendered (#959)
 
 ### Changed
 
--- a/docs/2.4/extensions/tables.md
+++ b/docs/2.4/extensions/tables.md
@@ -39,6 +39,11 @@ $config = [
             'tag' => 'div',
             'attributes' => [],
         ],
+        'alignment_attributes' => [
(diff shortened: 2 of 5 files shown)


colinodell:
@ptmkenny Would something like this work for you?

ptmkenny:
@colinodell Yeah, this is great-- it will eliminate the error for me and others who need to pass W3C for one reason or another.

Just a thought-- if we have the ability to inline CSS, is it also possible to add an option to supply the CSS classes for left/center/right align?  This might complicate the config too much, but many design frameworks (Bootstrap and Bulma are the ones I am familiar with, but I would assume others as well) already provide classes for doing this alignment, so being able to specify the classes to use would be the optimal solution from the perspective of a framework user.

ptmkenny:
@colinodell Wow, this looks great.

I've been thinking about it for the past couple days, and it seems robust enough to cover all the use cases I could think of--

* Using GFM as is
* Inlining the CSS to get W3C compliance
* Integrating with a framework
* Using arbitrary classes 

I definitely think the additional complexity is worth it because it empowers the user to make the extension compatible with whatever other systems they have in place.

ptmkenny:
@colinodell Thanks, I immediately installed and started testing this.

One snag-- when I use `class`, the rendered output includes both the result for `class` and `align`-- `align` doesn't get replaced; `class` just gets added.  The same happens for `style`-- `align` is also added.

In `TableCellRendered.php`, Xdebug shows that `$node->get->get('attributes')` has both `class` and `align` set, and `align` is set even if change the `DEFAULT_ATTRIBUTES` constant to be `class` values instead of `align`.  I was unable to debug further.

My understanding was that setting `class` or `style` is that it should replace the value for `align`, not supplement it, as the original goal was to generate W3C-compliant output, which requires removing `align`.

colinodell:
:facepalm: I forgot to instruct the configuration schema to replace (instead of merge) the default configs with the user-provided configs.  This should be fixed in 8721df81 - sorry about that!

ptmkenny:
@colinodell Thanks, with the latest version, it's working as expected!

PR #969, opened 2023-04-01
Change auto-prefixing to a secure protocol

GrahamCampbell:


GrahamCampbell:
Hmmm, github spec wants http. :/

colinodell:
We could introduce a config option for this

GrahamCampbell:
We could report this as a security vulnerability to GitHub. Surely this introduces a man in the middle vulnerability. If people need http, they should probably explicitly say it. ;)

colinodell:
I agree that `https` should be the default, and an issue has been opened on the GFM library proposing exactly that: https://github.com/github/cmark-gfm/issues/98

Unless/until the spec is updated accordingly, I think we should keep this library as-is, or at least make the default scheme configurable.

colinodell:
Closing this in favor of #987

PR #972, opened 2023-04-16
Fix links to `Mentions` extension documentation

faustbrian:
Currently those links 404 because they link to `mention` instead of `mentions`.


colinodell:
Thanks for catching and fixing these!

PR #976, opened 2023-05-09
fix: Home page examples since 2.0

joshbruce:
The example on the home page for 2.x didn't include:

1. a call to `getContent()` and
2. a comma in "Hello, World!"


colinodell:
Thanks for this!

(Technically the call to `getContent()` isn't necessary because the return value of `convert()` is `Stringable`, but I like that your change explicitly calls out the fact that you get an object back, thus avoiding the surprise)

joshbruce:
That is totally fair. I was bit by the surprise and was like, "Huh?" lol

Issue #985, opened 2023-07-29
Allow setting attributes without values in Attributes extension


svenluijten:
### Description

I'm writing my own extension that makes use of the capabilities of the baked-in attributes extension. However, upon testing, I found that it is currently not possible to add an attribute without a value on an element.

I would like to be able to add an attribute without any value to prevent me (and the eventual users of my own extension) from having to add an arbitrary value every time like so:

```md
![image](/assets/image.jpg){my-attribute=true}
```

### Example

Consider parsing the following Markdown with CommonMark, with the Attributes extension loaded:

```md
![image](/assets/image.jpg){my-attribute}
```

I would expect the following HTML to be rendered:

```html
<p><img src="/assets/image.jpg" alt="image" my-attribute /></p>
```

However, I'm seeing the following:

```html
<p><img src="/assets/image.jpg" alt="image" />{my-attribute}</p>
```

### Did this project help you today? Did it make you happy in any way?

I'm exceptionally happy with the way this project is set up. It was super easy to dig and start working on my own extension because of the outstanding documentation and the easy-to-follow code when I needed to dig deeper and do some source-diving. So thank you :heart

colinodell:
I really like this proposal! Let's do it :)

> I believe this has to do with the [League\CommonMark\Util\RegexHelper::PARTIAL_ATTRIBUTEVALUESPEC constant](https://github.com/thephpleague/commonmark/blob/5af699c47955345e14f66f3863ff7db48dcd90a2/src/Util/RegexHelper.php#L50) enforcing the presence of an equals character (=) and a subsequent value.

If possible, we should try to avoid changing those `PARTIAL_` regular expressions in `RegexHelper`.  Those are based directly on [the commonmark.js reference parser](https://github.com/commonmark/commonmark.js/blob/20b52e5bfe6729faccc0919026245da0f1b40ba3/lib/common.js#L23) and are used elsewhere in the library; any changes could have unintended side effects on other features like parsing HTML content.  Instead, we should be able to modify how they are used in `AttributesHelper::SINGLE_ATTRIBUTE`.

That regular expression basically says to match:

 - A CSS class or ID (`.` or `#` followed by some characters); or,
 - An attribute name followed by an "attribute value spec" (`=` then a value)

[...]

PR #986, opened 2023-07-29
Allow adding attributes without value

svenluijten:
As said in #985, this is my attempt at allowing valueless attributes to be set on elements.

### DIFF
--- a/src/Extension/Attributes/Util/AttributesHelper.php
+++ b/src/Extension/Attributes/Util/AttributesHelper.php
@@ -72,12 +72,14 @@ final class AttributesHelper
                 continue;
             }
 
-            if (\strpos($attribute, '=') === false) {
-                $attribute .= '=""';
+            $parts = \explode('=', $attribute, 2);
+            if (\count($parts) === 1) {
+                $attributes[$attribute] = true;
+                continue;
             }
 
             /** @psalm-suppress PossiblyUndefinedArrayOffset */
-            [$name, $value] = \explode('=', $attribute, 2);
+            [$name, $value] = $parts;
 
             $first = $value[0];
             $last  = \substr($value, -1);
--- a/src/Extension/Attributes/Util/AttributesHelper.php
+++ b/src/Extension/Attributes/Util/AttributesHelper.php
@@ -23,7 +23,7 @@ use League\CommonMark\Util\RegexHelper;
  */
 final class AttributesHelper
 {
-    private const SINGLE_ATTRIBUTE = '\s*([.#][_a-z0-9-]+|' . RegexHelper::PARTIAL_ATTRIBUTENAME . RegexHelper::PARTIAL_ATTRIBUTEVALUESPEC . ')\s*';
+    private const SINGLE_ATTRIBUTE = '\s*([.#][_a-z0-9-]+|' . RegexHelper::PARTIAL_ATTRIBUTENAME . RegexHelper::PARTIAL_ATTRIBUTEVALUESPEC . '?)\s*';
     private const ATTRIBUTE_LIST   = '/^{:?(' . self::SINGLE_ATTRIBUTE . ')+}/i';
 
     /**
--- a/src/Extension/Attributes/Util/AttributesHelper.php
+++ b/src/Extension/Attributes/Util/AttributesHelper.php
@@ -72,6 +72,10 @@ final class AttributesHelper
                 continue;
             }
 
+            if (\strpos($attribute, '=') === false) {
+                $attribute .= '=""';
+            }
+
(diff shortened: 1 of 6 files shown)


svenluijten:
Unfortunately the tests seem to fail as it is, probably because the regex is now too loose, causing _a single_ functional test to fail; one that has to do with illegal whitespace in HTML. I'd appreciate some help to fix that 🙂

colinodell:
Thanks for working on this! I [left some pointers](https://github.com/thephpleague/commonmark/issues/985#issuecomment-1658348484) on your original proposal which should hopefully help here.

src/Util/RegexHelper.php
```diff
@@ -47,7 +47,7 @@ final class RegexHelper
     public const PARTIAL_SINGLEQUOTEDVALUE     = '\'[^\']*\'';
     public const PARTIAL_DOUBLEQUOTEDVALUE     = '"[^"]*"';
     public const PARTIAL_ATTRIBUTEVALUE        = '(?:' . self::PARTIAL_UNQUOTEDVALUE . '|' . self::PARTIAL_SINGLEQUOTEDVALUE . '|' . self::PARTIAL_DOUBLEQUOTEDVALUE . ')';
-    public const PARTIAL_ATTRIBUTEVALUESPEC    = '(?:' . '\s*=' . '\s*' . self::PARTIAL_ATTRIBUTEVALUE . ')';
+    public const PARTIAL_ATTRIBUTEVALUESPEC    = '(?:' . '\s*=?' . '\s*' . self::PARTIAL_ATTRIBUTEVALUE . '?)';
```
colinodell, line 50: Let's keep this as-is.  But, when we reference this fragment, we can add `?` to the end to make this entire group optional.

src/Extension/Attributes/Util/AttributesHelper.php
```diff
@@ -72,6 +72,10 @@ public static function parseAttributes(Cursor $cursor): array
                 continue;
             }
 
+            if (\strpos($attribute, '=') === false) {
+                $attribute .= '=""';
+            }
+
             /** @psalm-suppress PossiblyUndefinedArrayOffset */
             [$name, $value] = \explode('=', $attribute, 2);
```
colinodell, line 80: The code below this point assumes we have a string value.  But if no value is given, we can skip that whole section with something like this:

```php
$parts = \explode('=', $attribute, 2);
if (\count($parts) === 1) {
    $attributes[$attribute] = true;
    continue;
}

[$name, $value] = $parts;
```

tests/unit/Extension/Attributes/Util/AttributesHelperTest.php
```diff
@@ -58,6 +58,7 @@ public function dataForTestParseAttributes(): iterable
         yield [new Cursor('{: target=_blank}'), ['target' => '_blank']];
         yield [new Cursor('{: target=_blank }'), ['target' => '_blank']];
         yield [new Cursor('{:   target=_blank   }'), ['target' => '_blank']];
+        yield [new Cursor('{: disabled}'), ['disabled' => '']];
```
colinodell, line 61: I think we probably want the attribute value to be `true` so that `HtmlElement` renders just the attribute name:

https://github.com/thephpleague/commonmark/blob/5af699c47955345e14f66f3863ff7db48dcd90a2/src/Util/HtmlElement.php#L124-L132

Could we also edit the test on line 56 to include a single attribute?  (That test case proves that different styles of attributes can be used together)

colinodell (changes requested):

svenluijten:
Thanks for the pointers! I'll work them out later today. Meanwhile, do you want me to point this PR to `main`? Seeing as you added [the issue](https://github.com/thephpleague/commonmark/issues/985) to the 2.5 milestone?

colinodell (approved):
LGTM!

svenluijten:
FWIW, [this](https://github.com/svenluijten/commonmark-image-media-queries) is what I was working on when I thought I could use valueless attributes 🙂

jasonvarga:
Hello. This PR has unfortunately caused some issues for us.

Apologies for commenting on an old PR, but this has only just been tagged today.

I'm happy to try and open a PR to fix this myself, but I thought I'd mention it here first as you may understand the problem better.

When we have templating language within our markdown string, previously the markdown parser would leave it alone. Now, it basically clears it out. For example:

```php
$converter->convert('# Hello {{ foo }}');

// Before: <h1>Hello {{ foo }}</h1>
// After: <h1>Hello {}</h1>
```

We use the AttributesExtension. If I revert the `SINGLE_ATTRIBUTE` const change from this PR, it works.

PR #988, opened 2023-08-02
Replace symbol with backref_symbol

clustermin:
It should be `backref_symbol`, not `symbol`.


colinodell (approved):
Thank you for catching this!

PR #996, opened 2023-09-20
Replace the array_merge with spread array operator

peter279k:
# Changed log

- According to the [reference](https://wiki.php.net/rfc/spread_operator_for_array), it's great to use the spread array operator to replace the `array_merge` function.
- There're two reasons about using the spread array operator:
  - Spread operator should have a better performance than array_merge and compile time optimization can be performant for constant arrays.
  - `array_merge` only supports array, while the spread operator also supports objects implementing `Traversable`.


colinodell (approved):
Thanks for the contribution!

PR #997, opened 2023-10-09
Test against php 8.2 and 8.3

sergiy-petrov:
* Test against php 8.3

* Test against php 8.2 in stable mode


colinodell (approved):
Thank you for this!

PR #1011, opened 2024-02-02
CommonMark spec 0.31 compatibility

colinodell:
This PR brings our library into compliance with [the latest version (0.31) of the CommonMark spec](https://spec.commonmark.org).


colinodell:
BC checks are failing because the values of some constants are changing.  These particular constants contain regular expressions used to parse Markdown.  Because the constants remain usable for the same purposes, we consider this to not be a BC-break.

PR #1013, opened 2024-02-08
Allow CSS level 3 selectors for classes and ids in the attributes extension

xavierlacot:
The [CSS level 3 Syntax](https://www.w3.org/TR/selectors-3/), which is a Recommendation since 2018, allows much more characters in `id` and `class` selectors than the ones currently supported by the "Attributes" extension:

https://github.com/thephpleague/commonmark/blob/91c24291965bd6d7c46c46a12ba7492f83b1cadf/src/Extension/Attributes/Util/AttributesHelper.php#L26

The formal CSS level 3 selectors grammar is detailed here: https://www.w3.org/TR/selectors-3/#w3cselgrammar

This pull request adds a test to check several valid `class` and `id` names that are currently failing to be parsed as valid attribute values by the "Attributes" extension, and it proposes a fix for this issue.

### DIFF
--- a/src/Extension/Attributes/Util/AttributesHelper.php
+++ b/src/Extension/Attributes/Util/AttributesHelper.php
@@ -23,7 +23,7 @@ use League\CommonMark\Util\RegexHelper;
  */
 final class AttributesHelper
 {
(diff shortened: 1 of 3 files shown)


colinodell (approved):
Thank you for contributing this improvement!

PR #1016, opened 2024-03-21
Update embed.md

nfreader:
Changes an errant instance of `composer install` to `composer require` in the docs.


colinodell (approved):
Thank you for catching and fixing this!

PR #1025, opened 2024-05-22
UrlAutolinkParser fails for text containing WWW and a subsequent link

macbookandrew:
For strings containing `www` followed by a link using `[link](https://…)` format, the parser mangles the resulting HTML:

```html
<!-- given this string of markdown -->
WWW text followed by a [link](https://example.com/foo-bar-test)

<!-- expected result -->
<p>WWW text followed by a <a href="https://example.com/foo-bar-test">link</a></p>

<!-- actual result -->
<p><a href="https://example.com/foo-bar-test">https://example.com/foo-bar-test</a>tps://example.com/foo-bar-test)</p>
```


colinodell (approved):
Thank you for spotting this issue and fixing it!

PR #1040, opened 2024-08-13
Fix the parser behavior for texts containing curly-braces when the Attributes extension is enabled

xavierlacot:
## The issue

As stated in https://github.com/thephpleague/commonmark/pull/1035#issuecomment-2284664894, the PR #986 , which introduced the ability to create valueless attributes using the Attributes extension, causes issues when a markdown document includes text strings that contain curly braces.

For example, the following markdown text

```md
Elastic{ON} Tour San Francisco
```

is rendered by [the official commonmark dingus](https://spec.commonmark.org/dingus/) as:

```html
<p>Elastic{ON} Tour San Francisco</p>
```

The `league/commonmark` library with the "Attributes" extension enabled was used to render such markdown the same way until the commit 2138460860e2cee78498846a9d6e2da1913bbadb. Since then, it renders this markdown as:

```html
<p>Elastic Tour San Francisco</p>
```

which is obviously something we do not want.

## What does this PR

This PR reverts the changes introduced in #986 and  #1035 and adds more test cases, to show how to create valid [boolean-attributes](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes).

A section has also been added in the documentation to show how to write attributes that will be rendered as empty-value attributes in HTML.

### DIFF
--- a/docs/2.5/extensions/attributes.md
+++ b/docs/2.5/extensions/attributes.md
@@ -55,6 +55,21 @@ Output:
 <p>This is <em style="color: red">red</em>.</p>
 ```
 
+### Empty-Value Attributes
+
+Attributes can be rendered in HTML without a value by using `true` value in the markdown document:
+
+```markdown
+{itemscope=true}
+## Header
+```
+
+Output:
+
+```html
+<h2 itemscope>Header</h2>
+```
+
 ## Installation
 
 This extension is bundled with `league/commonmark`. This library can be installed via Composer:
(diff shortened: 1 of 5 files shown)


colinodell (approved):
Thank you for the PR!

While I do like the simplicity of allowing `{attribute}`, I think it's entirely reasonable to revert that feature in favor of the more explicit approach that is less likely to collide with other user input.

PR #1044, opened 2024-08-26
Update README to list PHP 8 as required

nickmendoza:
nette/utils requires PHP >= 8.0 < 8.4, so PHP 7.4 will no longer pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 ## 📦 Installation & Basic Usage
 
-This project requires PHP 7.4 or higher with the `mbstring` extension.  To install it via [Composer] simply run:
+This project requires PHP 8.0 or higher with the `mbstring` extension.  To install it via [Composer] simply run:
 
 ``` bash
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I'm not sure that's right. PHP 7.4 is fully supported by this package.

nickmendoza:
Looks like if you force in a version of it that is not 4.0, then you're right.
```
│  │  ├──nette/utils ^2.5.7 || ^3.1.5 ||  ^4.0
│  │  │  └──php >=8.0 <8.4
```

GrahamCampbell:
You don't need to force it, as the CI demonstrates?

colinodell:
Composer will automatically select the newest version that is compatible with your local PHP version. For example, on PHP 7.4 you'll get v3.2.10:  https://github.com/thephpleague/commonmark/actions/runs/10419410435/job/28857397686#step:5:45

PR #1048, opened 2024-09-03
Footnote documentation - add example

edent:
The current example didn't contain a footnote to test. Anyone copying and pasting the example would have been confused (I know I was!)


colinodell (approved):
This is great, thank you!

PR #1050, opened 2024-09-07
Fixed config typo in autolinks.md

ramondelafuente:
The current docs result in an error:
```
[League\Config\Exception\ValidationException] Unexpected item 'autolink › default_protocols', did you mean 'default_protocol'? 
```


colinodell (approved):
Thanks for the fix!

PR #1052, opened 2024-09-24
Make it possible to add extension from the CommonMarkConverter class

Pierstoval:
In my use-case, the application is using CommonMark in a non-usual way, and I just need to add extensions to it **and** append configuration to the Environment (it's in a Symfony context).

However, considering the current architecture, there are only two possibilities for the package in Symfony DI's setup:

* Use the `LeagueCommonMarkConverterFactory` from TwigExtraBundle. You can add extensions (they're injected in the constructor, and services can just be tagged, but you cannot add any configuration to the `array $config` constructor parameter.
* Create a standard DI's `Definition` with any DI way (I'm using Yaml, but all other formats are concerned). This allows adding custom configuration to the constructor. But to add extensions, you have to call `$converter->getEnvironment()->addExtension()`, which is not possible with Symfony's DI.

For now, there is no way to do it, either from TwigExtra or from CommonMark itself.

I will also make a PR on TwigExtra to allow better customization of the factory itself, for more flexibility

### DIFF
--- a/src/CommonMarkConverter.php
+++ b/src/CommonMarkConverter.php
@@ -18,6 +18,7 @@ namespace League\CommonMark;
 
 use League\CommonMark\Environment\Environment;
 use League\CommonMark\Extension\CommonMark\CommonMarkCoreExtension;
+use League\CommonMark\Extension\ExtensionInterface;
 
 /**
  * Converts CommonMark-compatible Markdown to HTML.
@@ -42,5 +43,10 @@ final class CommonMarkConverter extends MarkdownConverter
         \assert($this->environment instanceof Environment);
 
         return $this->environment;
+    }
+
+    public function addExtension(ExtensionInterface $extension): void
+    {
+        $this->environment->addExtension($extension);
     }
 }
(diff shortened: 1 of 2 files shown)


Pierstoval:
@colinodell The problem in that case mostly comes from TwigExtra, since the `Twig\Extra\Markdown\LeagueMarkdown` class depends specifically on `League\CommonMark\CommonMarkConverter`.

I can indeed create a definition for the `League\CommonMark\Environment\Environment` class, but as said in the first post, there is no way to inject it in `League\CommonMark\CommonMarkConverter`. In this PR I had only two choices: allow adding extensions from the Converter (as a wrapper), or add an optional `Environment` parameter to the constructor. Since the config is already there, I decided it was easier to execute `->addExtension` later (since Symfony DI allows it)

colinodell:
Hey, so sorry for taking over a year to respond to this 🤦

The `CommonMarkConverter` is specifically designed for simple cases where developers only want the `CommonMarkCoreExtension` and nothing else. Think of it like "easy mode" for anyone who doesn't care about additional/custom syntax.

So while I do appreciate you taking the time to put this together, I think the better approach would be for TwigExtra to accept a `ConverterInterface` instance instead.

PR #1055, opened 2024-12-03
Remove redundant code

chillbram:
This PR removes a call to `min()` for two values that by virtue of the previous if statement must already be exactly identical. As they are always the same we can remove both the call to `min()` and one of the two `->getLength()` calls.


colinodell:
Thanks for spotting this! Would you mind adding a comment like:

```
// Both are the same length so just return one of them
```

Something to remind me to not re-add silly logic again haha

chillbram:
Added a little comment :+1:

colinodell (approved):
Perfect, thank you so much!

PR #1063, opened 2025-01-25
fix incorrect interface in docs v2.6

CharrafiMed:
just updating the docs for prevent confusing for other developers when trying to build custom extensions


colinodell (approved):

PR #1066, opened 2025-02-23
Add missing newline in Front Matter documentation example

elazar:
There's a missing newline between imports in one of the documentation examples for the Front Matter extension that affects its readability.


colinodell (approved):
Thanks for fixing this!

PR #1074, opened 2025-04-30
[TableOfContents] Add a label option

DanielEScherzer:
By default no label is shown, but if one is set it is added as a `<strong>` before the first item in the table of contents.

### DIFF
--- a/src/Extension/TableOfContents/TableOfContentsBuilder.php
+++ b/src/Extension/TableOfContents/TableOfContentsBuilder.php
@@ -43,6 +43,7 @@ final class TableOfContentsBuilder implements ConfigurationAwareInterface
             (int) $this->config->get('table_of_contents/min_heading_level'),
             (int) $this->config->get('table_of_contents/max_heading_level'),
             (string) $this->config->get('heading_permalink/fragment_prefix'),
+            (string) $this->config->get('table_of_contents/label'),
         );
 
         $toc = $generator->generate($document);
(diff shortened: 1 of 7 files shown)


DanielEScherzer:
The failing check `continuous-integration/styleci/pr` says
> We were unable to access the repo or commit to analyze it.

So I don't think that is something that I can fix

froschdesign:
> …it is added as a `<strong>` before the first item in the table of contents.

Which produces invalid HTML: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ul#technical_summary

colinodell:
Have you considered using the `'position' => 'placeholder'` option instead? It allows much more flexibility, such as providing a heading (which could be made linkable thanks to [the Heading Permalinks extension](https://commonmark.thephpleague.com/2.7/extensions/heading-permalinks/)):

```md
## Table of Contents

[TOC]
```

DanielEScherzer:
> Have you considered using the `'position' => 'placeholder'` option instead? It allows much more flexibility, such as providing a heading (which could be made linkable thanks to [the Heading Permalinks extension](https://commonmark.thephpleague.com/2.7/extensions/heading-permalinks/)):
> 
> ```
> ## Table of Contents
> 
> [TOC]
> ```

I wanted the heading to be more clearly part of the TOC itself, so that is why I didn't have a placeholder.
You can see how I'm using a label at https://scherzer.dev/Blog/20250429-attributes-on-constants in the sidebar 

> > …it is added as a `\<strong\>` before the first item in the table of contents.
> 
> Which produces invalid HTML: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ul#technical_summary

Good point - what would you think of having a wrapping div when a label is requested?
```html
<div class="table-of-contents-wrapper">
<strong>Table of Contents</strong>
<ul class="table-of-contents">
<li>...
</ul>
</div>
```

PR #1081, opened 2025-05-25
Update EmbedProcessor.php

driesvints:
Prevent subsequent calls if there's no embeds.


colinodell (approved):
Thanks for catching and contributing this!

PR #1091, opened 2025-10-10
Update HeadingPermalinkRenderer.php

filerun:
Added slug delimiter config. I needed slugs to use underscores instead of dashes. I could have extended the SlugNormalizer to do a str_replace as a quick solution for my needs, but the class is final. The library could use this option. (I edited the files in the browser, hence the patches rather than a single pull request. I hope it is going to be useful to somebody.)


colinodell:
See https://github.com/thephpleague/commonmark/pull/1093#issuecomment-3581863724

PR #1096, opened 2025-10-31
case insensitive for HTML attributes

RobinDev:
The HTML parser treated attribute names in a case-sensitive manner, which is not compliant with the HTML specifications.

This behavior caused incorrect escaping of tags containing attributes with uppercase letters, such as the `viewBox` attribute in SVG tags.

This change updates the `PARTIAL_ATTRIBUTENAME` regular expression to accept both uppercase and lowercase letters, ensuring that HTML attributes are correctly recognized regardless of their case.


RobinDev:
Input:

```markdown
Icon: <svg viewBox="0 0 4 4"><path d="M0 0h4v4z"/></svg> end.
```

Output (2.10.0, default `CommonMarkConverter`):

```html
<p>Icon: &lt;svg viewBox=&quot;0 0 4 4&quot;&gt;<path d="M0 0h4v4z"/></svg> end.</p>
```

The uppercase letter in `viewBox` keeps the tag from matching inline raw HTML, so its `<` is escaped — `<path>` (lowercase attributes) passes through. The spec allows it: an attribute name is "an ASCII letter, `_`, or `:`, followed by zero or more ASCII letters, digits, …" ([§6.6](https://spec.commonmark.org/0.31.2/#attribute-name)); commonmark.js uses `attributename = "[a-zA-Z_:][a-zA-Z0-9:._-]*"` ([lib/common.js](https://github.com/commonmark/commonmark.js/blob/master/lib/common.js)) — [dingus](https://spec.commonmark.org/dingus/?text=Icon%3A%20%3Csvg%20viewBox%3D%220%200%204%204%22%3E%3Cpath%20d%3D%22M0%200h4v4z%22%2F%3E%3C%2Fsvg%3E%20end.).

Only inline is affected: the type-7 HTML block start regex is compiled with `/i`, so the same tag alone on its own line already works. This PR aligns inline with blocks.

Branch rebased on 2.7; the example is now a `LocalDataTest` fixture (`raw_html_uppercase_attributes`) plus two `RegexHelperTest` assertions.

colinodell:
Thank you!

PR #1100, opened 2025-11-25
Add Mark (highlight) extension

ossobuffo:
This is a feature request to add a highlight extension to Commonmark (see [the Markdown Guide's documentation](https://www.markdownguide.org/extended-syntax/#highlight) on this feature). This extension is not loaded by default, but is available for implementations that want to support highlighting.

This was quite straightforward, since it merely involved copying and modifying the Strikethrough extension and its tests.

The one unanswered question I have is whether `mark` is the correct return value for `getXmlTagName()` for this extension.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
 
 ## [Unreleased][unreleased]
 
+### Added
+- Added a new `HighlightExtension` for marking important text using `==` syntax (#1100)
+
 ## [2.7.0]
 
 This is a **security release** to address a potential cross-site scripting (XSS) vulnerability when using the `AttributesExtension` with untrusted user input.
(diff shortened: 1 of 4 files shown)


colinodell:
Thanks for suggesting this!

Although I've been hesitant to include new extensions in this library, I'm inclined to include this one because:

- The implementation is small, making the maintenance burden trivial
- Most (all?) parsers that do support highlights seem to use this syntax

> The one unanswered question I have is whether `mark` is the correct return value for getXmlTagName() for this extension.

Yep, that works for me!

There are a few small changes I think we should make:

- Rename the extension itself to `Highlight`, but continue using `Mark` for the inline element type
- Fix the failing pipelines (code style, and copypaste errors in tests)
- Add this to the docs

I've got some free time now, so I'll give that a go :)

colinodell:
Thanks for contributing this! I'll include this in the next minor release (v2.8.0)

PR #1107, opened 2026-01-09
Fix PHP 8.5 deprecation

freost:
Passing strings of more than one byte to ord() is deprecated in PHP 8.5. This change ensures that ord() is only called if the string is one byte.


colinodell:
Thanks for the contribution!

f1mishutka:
Any chance this to be released soon (2.8.1 probably?).
PHP 8.5 released more then month ago so this fix is quite critical...

Thanks!

PR #1108, opened 2026-01-12
Fix boolean default value in tag method

AJenbo:
It's not technically wrong but it's different then in other parts of the code and some parsers struggle with it.


colinodell:
Thanks!

PR #1109, opened 2026-01-22
perf(core): Promote Text delimiter flag to property

### LINKED ISSUE #963: Promote 'delim' to property
Instead of using `Text::$data` to store whether it might be a delimiter, create a new `$isDelimiter` property. This should speed up delimiter processing by ~10%.

This will be a breaking change.  Also, update the docs accordingly.

SilasArche:
Closes #963

Addresses issue #963.

Previously, the `InlineParserEngine` checked `$node->data->has('delim')` every time it attempted to merge adjacent text nodes. This resulted in frequent array lookups in the hot path of the parser.

This commit promotes the delimiter status to a first-class property on the `Text` node. The constructor now strips the `delim` key from the `$data` array and sets the boolean property `$isDelimiter` instead.

This reduces overhead for text merging operations without changing the public API surface for extension authors who instantiate `Text` nodes.

"We polish the silver." - The Wake Guild

### DIFF
--- a/src/Node/Inline/Text.php
+++ b/src/Node/Inline/Text.php
@@ -18,6 +18,17 @@ namespace League\CommonMark\Node\Inline;
 
 final class Text extends AbstractStringContainer
 {
+    public bool $isDelimiter = false;
+
+    /**
+     * @param string               $contents The text contents
+     * @param array<string, mixed> $data     Arbitrary data
+     */
+    public function __construct(string $contents = '', array $data = [])
+    {
+        parent::__construct($contents, $data);
+    }
+
     public function append(string $literal): void
     {
         $this->literal .= $literal;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
 
 ## [Unreleased][unreleased]
 
+### Changed
+- Promoted the `delim` status on `Text` nodes to a first-class property (`$isDelimiter`) instead of storing it in the `$data` array (#963).
+    - This improves performance in the inline parser engine.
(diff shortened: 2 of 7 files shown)


src/Node/Inline/Text.php
```diff
@@ -18,6 +18,32 @@
 
 final class Text extends AbstractStringContainer
 {
+    private bool $isDelimiter = false;
```
colinodell: Let's just make this `public` so we don't need getter and setter functions.

src/Node/Inline/Text.php
```diff
@@ -18,6 +18,32 @@
...
+    /**
+     * @param string               $contents The text contents
+     * @param array<string, mixed> $data     Arbitrary data
+     */
+    public function __construct(string $contents = '', array $data = [])
+    {
+        if (isset($data['delim'])) {
+            $this->isDelimiter = (bool) $data['delim'];
+            unset($data['delim']);
+        }
```
colinodell: Because this MR introduces a breaking change, we'll need to bump the library's major version, making this a good opportunity to incorporate other breaking changes too. So I think we should completely drop support for setting/passing `delim` via the data array everywhere.

colinodell (changes requested):
Thanks for working on this!

I think I'd prefer to fully switch to using the new property, though - no getters, setters, or lingering support for the `delim` property anywhere in the codebase. Would you be open to making those changes?

src/Node/Inline/Text.php
SilasArche: Agreed. It's a more direct approach and avoids the method call overhead. I've made the property public and removed the getter/setter methods.

src/Node/Inline/Text.php
SilasArche: That's a good point. No sense in carrying legacy code through a major version bump. I've removed the backwards-compatibility check from the constructor and updated the DelimiterParser to set the public property directly.

PR #1112, opened 2026-02-13
Allow disabling of inline footnotes + documentation on inline footnotes

chillbram:
Problem: while using our new [sup extension](https://github.com/Cradeq/commonmark-strikethrough-sub-sup-extension), adding a link to superscript (`^[Text](link)^`) suddenly created footnotes.

I couldn't find any documentation on the `^[Footnote]` notation, but diving in the code base I found out this was a feature. So I added the explanation of the feature I found in the test suite and also added some example code to the regular docs.

To fix the incompatibility, I propose this new config value `enable_inline_footnotes` to allow users to disable inline footnotes. The implementation should be fully backwards compatible, as the default is still set to true. I added a test for this behavior.


colinodell (approved):
Thanks for the contribution!

PR #1113, opened 2026-02-24
Add NormalizeHeadingsExtension to constrain heading levels

### LINKED ISSUE #989: Heading levels
### Description

Please, add ability to control allowed heading levels such as [CKEditor do](https://ckeditor.com/docs/ckeditor5/latest/features/headings.html#heading-levels)

### Example

May be via env config, similar to heading_permalink:

```php
'heading_permalink' => [
    'min_heading_level' => 1,
    'max_heading_level' => 6,
],
```

To:

```php
'headings' => [
    'min_level' => 2,
    'max_level' => 6,
],
```

### Did this project help you today? Did it make you happy in any way?

Definitely love it ❤️

eyupcanakman:
This adds a new `NormalizeHeadingsExtension` to constrain heading levels to a configured min/max range.

For example, with:

```php
'normalize_headings' => [
    'min_level' => 2,
    'max_level' => 4,
],
```

`#` headings are rendered as `<h2>`, and `#####` / `######` headings are rendered as `<h4>`.

The extension observes `DocumentParsedEvent` and clamps heading nodes in the parsed AST. Configuration now validates both bounds (`1..6`) and the `min_level <= max_level` invariant.

Also included:
- unit and functional tests (including HeadingPermalink + TableOfContents interoperability)
- extension docs page + overview/menu updates
- changelog entry

Fixes #989


colinodell:
Thanks for the contribution!

I've decided to rename some things so we can bundle #1115 into the same extension, but I kept your implementation as-is :)

colinodell (approved):

samwilson:
I added [a comment](https://github.com/thephpleague/commonmark/issues/989#issuecomment-4150026311) to the task about how this makes it possible to end up with duplicate headings. Is that a concern at all do you think?

colinodell:
> I added [a comment](https://github.com/thephpleague/commonmark/issues/989#issuecomment-4150026311) to the task about how this makes it possible to end up with duplicate headings. Is that a concern at all do you think?

Yeah, now that I think about it, the clamp-only behavior feels like a default nobody will really want. We should probably renumber by default - similar to #1115 which I'm writing code for now.  Let me bake this idea into that :)

colinodell:
Done in #1127 — renumbering is now the extension's default, so your `min_level: 3` example renders `<h3>Foo</h3><h4>Bar</h4>` instead of two `<h3>`s. That should cover the standalone-vs-subsection case you raised in #989 as well.

tomsommer:
Thank you for this, just what I was looking for 👍

PR #1119, opened 2026-05-13
Optimize Cursor: skip strrpos on tab-free documents

GromNaN:
`Cursor::__construct()` called `strrpos($line, "\t")` on every line to locate the last tab position. For tab-free documents this is a full backward scan for no benefit — 40–165 ns/line depending on length.

`MarkdownParser::parse()` now scans the document once with `str_contains()` (~3 µs). When no tabs are found, `new Cursor($line, false)` sets `lastTabPosition = false` directly without calling `strrpos`. The `$lineCouldHaveTabs` parameter defaults to `true`, keeping the public API backward compatible.

Benchmarks run with `tests/benchmark/benchmark_parse.php` (new warm script: converter initialized once, 100 iterations + 10 warmup, `hrtime()`). PHP 8.5.2, OPcache on, Xdebug off.

`sample.md` (27 KB, 2.9% tabs):

| | median | p95 |
|---|---|---|
| Before | 4.38 ms | 6.26 ms |
| After  | 3.92 ms | 4.94 ms |

Larger corpora from vendor (not committed): CHANGELOG.md 570 KB −5.2%, spec.txt 200 KB −2.8%.

### DIFF
--- a/src/Parser/Cursor.php
+++ b/src/Parser/Cursor.php
@@ -59,7 +59,7 @@ class Cursor
     /**
      * @param string $line The line being parsed (ASCII or UTF-8)
      */
-    public function __construct(string $line)
+    public function __construct(string $line, bool $lineCouldHaveTabs = true)
     {
         if (! \mb_check_encoding($line, 'UTF-8')) {
             throw new UnexpectedEncodingException('Unexpected encoding - UTF-8 or ASCII was expected');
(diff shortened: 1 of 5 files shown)


colinodell:
Thank you for the suggestion! I tried replicating this benchmark on my machine and wasn't able to get more than a 0.6% improvement. While the optimization makes sense, I don't love that we'd need to expand the public interface for such a tiny improvement. Instead, I'd rather save this idea for #1142, where we might have a highly-optimized `TabFreeASCIICursor`.

PR #1126, opened 2026-07-12
Add opt-in literal default attribute strings

### LINKED ISSUE #1123: Link can't have link class
### Version(s) affected

2.8.2

### Description

            'default_attributes' => [
                Link::class => [
                    'class' => "link",
                ],

this will not work and will give a

> **link() expects exactly 2 arguments, 1 given**

if i use link-1 it works, i tried btn btn-link and it worked

### How to reproduce

```
<?php
namespace App\Services;

use League\CommonMark\Extension\CommonMark\CommonMarkCoreExtension;
use League\CommonMark\Extension\CommonMark\Node\Inline\Image;
use League\CommonMark\Extension\CommonMark\Node\Inline\Link;
use League\CommonMark\Extension\DefaultAttributes\DefaultAttributesExtension;
use League\CommonMark\Extension\GithubFlavoredMarkdownExtension;
use League\CommonMark\MarkdownConverter;

class Md2htmlService
{
    protected MarkdownConverter $converter;

    public function __construct()
    {
        $config = [
            'default_attributes' => [
                Link::class => [
                    'class' => "link", // TODO: issue lives here
                ],
                Image::class => [
                    'loading' => 'lazy',
                ]
            ]
        ];
        $environment = new Environment($config);
        $environment->addExtension(new CommonMarkCoreExtension());
        $environment->addExtension(new GithubFlavoredMarkdownExtension());

        $environment->addExtension(new DefaultAttributesExtension());
        

        $this->converter = new MarkdownConverter($environment);
    }

    public function convert(string $markdown): string
    {
        return $this->converter->convert($markdown)->getContent();
    }
}
```

oiahoon:
## Summary

- add a `default_attributes_literal_strings` option for treating callable strings and string arrays as literal attribute values
- preserve existing string and array callback behavior by default for backwards compatibility
- keep closures and other non-string callable types working when the option is enabled
- document the option, callable return values, and named callback wrapping

## Testing

- focused DefaultAttributes tests: 8 tests, 10 assertions
- full PHPUnit suite: 4,184 tests, 5,168 assertions
- pathological suite: 44 cases
- targeted PHPStan analysis: no errors
- PHP syntax and `git diff --check`

Closes #1123

### DIFF
--- a/docs/2.x/extensions/default-attributes.md
+++ b/docs/2.x/extensions/default-attributes.md
@@ -49,6 +49,7 @@ use League\CommonMark\Node\Block\Paragraph;
 // Extension defaults are shown below
 // If you're happy with the defaults, feel free to remove them from this array
 $config = [
+    'default_attributes_literal_strings' => false,
     'default_attributes' => [
         Heading::class => [
             'class' => static function (Heading $node) {
@@ -93,7 +94,11 @@ Attribute values may be any of the following types:
 - `string`
 - `string[]`
 - `bool`
-- `callable` (parameter is the `Node`, return value may be `string|string[]|bool`)
+- `callable` (parameter is the `Node`, return value may be `string|string[]|bool|null`)
+
+For backwards compatibility, strings and string arrays that PHP considers callable are invoked by default. Set `default_attributes_literal_strings` to `true` to always treat those values as literal attributes. Other callable types, such as closures and invokable objects, remain callable.
+
+When literal string values are enabled, wrap named function or method callbacks with `\Closure::fromCallable()`.
 
 ## Examples
 
--- a/src/Extension/DefaultAttributes/ApplyDefaultAttributesProcessor.php
+++ b/src/Extension/DefaultAttributes/ApplyDefaultAttributesProcessor.php
@@ -26,6 +26,7 @@ final class ApplyDefaultAttributesProcessor implements ConfigurationAwareInterfa
     {
         /** @var array<string, array<string, mixed>> $map */
         $map = $this->config->get('default_attributes');
+        $literalStrings = $this->config->get('default_attributes_literal_strings');
 
         // Don't bother iterating if no default attributes are configured
         if (! $map) {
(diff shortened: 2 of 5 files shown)


src/Extension/DefaultAttributes/ApplyDefaultAttributesProcessor.php
```diff
@@ -40,7 +40,7 @@ public function onDocumentParsed(DocumentParsedEvent $event): void
 
             $newAttributes = [];
             foreach ($attributesToApply as $name => $value) {
-                if (\is_callable($value)) {
+                if (! \is_string($value) && \is_callable($value)) {
```
colinodell: I think we may need similar treatment for arrays. For example, `['Locale', 'getDefault']` is also a valid `callable`.

docs/2.x/extensions/default-attributes.md
```diff
@@ -95,6 +95,8 @@ Attribute values may be any of the following types:
 - `bool`
 - `callable` (parameter is the `Node`, return value may be `string|string[]|bool`)
 
+String values are always treated as literal attribute values. To use a named function as a callback, wrap it with `Closure::fromCallable()`.
```
colinodell: Could we please include a test showing this in action? I'm sure it will work just fine - I just like having tests should we ever change this behavior in future versions.

colinodell (commented):
I agree that we need to do something about #1123, but this change (as-is) breaks backward-compatibility. Perhaps we should introduce a new boolean config option to opt into this fix? Open to any other ideas on how to minimize the BC break here.

src/Extension/DefaultAttributes/ApplyDefaultAttributesProcessor.php
oiahoon: Addressed in 956f5ee. The new option treats both callable strings and string arrays as literals, while the default preserves existing callback behavior.

docs/2.x/extensions/default-attributes.md
oiahoon: Addressed in 956f5ee. Focused coverage now exercises legacy string and array callbacks, opt-in literal values, and closure callbacks.

colinodell:
Thank you for implementing this! Ultimately, I think I want to take a _slightly_ different path via #1148 - it's basically the same idea, but with a different config structure and some other minor tweaks. Your PR was extremely helpful for me to understand some of the potential solutions and tradeoffs for solving this, so thank you so much for your contribution!

PR #1129, opened 2026-07-12
Strengthen unsafe link and image renderer test assertions

GrahamCampbell:
Just a quick PR to make some test assertions stricter and more consistent with sister tests.


colinodell (approved):
Thanks!

PR #1131, opened 2026-07-12
Fix over-blocking of safe URLs by the unsafe link filter

GrahamCampbell:
When `allow_unsafe_links` is disabled, the `^` anchor in `REGEX_UNSAFE_PROTOCOL` binds only to the first alternative, so `vbscript:`, `file:`, and `data:` match anywhere in the URL rather than just at the start. Safe links such as https://example.com/view?src=data:image/png are misclassified as unsafe and silently lose their `href/src`.

The first commit adds tests demonstrating the bug and the second applies the one-line fix.

---

The JS implementation has a similar bug: https://github.com/commonmark/commonmark.js/pull/305.


colinodell (approved):

GrahamCampbell:
Nice, this was merged in the JS package too, so they are kept in sync. :)

PR #1134, opened 2026-07-12
Render the TOC once and share it across placeholders

GrahamCampbell:
When the TableOfContents extension runs in placeholder mode, every placeholder currently receives its own deep clone of the generated TOC, so a document with `P` placeholders and `H` headings does `O(P × H)` cloning and rendering work. This change replaces each placeholder with a lightweight `TableOfContentsReference` that points at a single shared TOC, which is rendered once and reused — the HTML output stays byte-for-byte identical (the existing fixtures pin this), while the AST and render work drop to `O(P + H)`. Non-HTML formats like XML still see complete copies via a pre-render expansion, so XML output is unchanged too.

There's also a new opt-in `table_of_contents/max_toc_entries` option (default unlimited) that caps the total number of rendered TOC entries for anyone feeding placeholder mode untrusted input; once the budget is spent, remaining placeholders are simply left unrendered.

---

Some quick benchmarks (PHP 8.4, `P` placeholders × `H` headings, identical output confirmed by hash):

| P = H | Time before | Time after | Peak memory before | Peak memory after | Output size |
|------:|------------:|-----------:|-------------------:|------------------:|------------:|
|    50 |      0.022s |     0.009s |             6.0 MB |            4.0 MB |     0.14 MB |
|   100 |      0.066s |     0.009s |            12.0 MB |            6.0 MB |     0.53 MB |
|   200 |      0.251s |     0.012s |            34.3 MB |            8.3 MB |     2.13 MB |
|   400 |      1.003s |     0.021s |           121.1 MB |           23.1 MB |     8.53 MB |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,16 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
 
 ## [Unreleased][unreleased]
 
+### Added
+- Added a new `table_of_contents/max_placeholder_entries` option to limit how many table of contents entries a document may render across all of its placeholders (#1134)
+
 ### Changed
 - Improved performance of reading single characters from multibyte lines
 - Improved performance of locating the next non-space character on lines without tabs
 - Optimized `Cursor::advanceToNextNonSpaceOrNewline()` to scan the line in place instead of copying everything left in the block on every call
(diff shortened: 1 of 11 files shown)


GrahamCampbell:
Squashed and rebased, including your commits @colinodell.

colinodell:
> Squashed and rebased, including your commits @colinodell.

Thank you! I'll get this merged shortly.

PR #1136, opened 2026-07-18
Cap the default block nesting level at 100

### LINKED ISSUE #1137: Make default options safe by default
v1 and v2 of this library have always prioritized parsing correctness, especially for the default configuration values. These options are completely reasonable for trusted inputs, but can lead to vulnerabilities (especially denial-of-service) when handling untrusted inputs from potentially malicious users. Our README, documentation, etc. all try to make these risks and mitigations well-known: https://commonmark.thephpleague.com/2.x/security/

For v3, I think we should change to **safe by default**, where reasonable default limits and security settings come out-of-the-box unless explicitly configured otherwise.

## Proposed Implementation

In 2.x, we should use `@trigger_error()` to notify whenever an insecure default config is used. This would hopefully encourage users to start providing the explicit values they want instead of accepting library defaults.

In 3.x, we would either:

1. Change all default values for security-related options to reasonable defaults; or,
2. Remove the default values so that users must provide their own explicit values
    - We could also provide defaults as presets with names like `safe` or `unsafe`

GrahamCampbell:
This is the behavior-changing alternative to #1135. It caps nesting at 100 by default, preserves intentionally deeper fixtures with explicit overrides, and documents the compatibility trade-off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
 - Improved performance of locating the next non-space character on lines without tabs
 - Optimized `Cursor::advanceToNextNonSpaceOrNewline()` to scan the line in place instead of copying everything left in the block on every call
 - Optimized inline link destination parsing to scan the line in place, so its cost follows the length of the destination rather than the length of everything left in the block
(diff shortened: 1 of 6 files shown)


colinodell:
Let's tackle this as part of #1137 instead

PR #1144, opened 2026-08-10
Fix Cursor::match() treating text before the cursor as part of the subject

colinodell:
### What broke

Since 2.9.0, `Cursor::match()` runs the pattern against the whole persistent line at the cursor's byte offset, rewriting a leading `^` to `\G`. That rewrite covers the common case, but PCRE keeps the text *before* a start offset visible to the pattern, so any construct that inspects what precedes the match start silently changed meaning. With the cursor advanced by 1 on `"abcdef"` (`/m` case on `"ab\ncd"`):

| pattern | 2.8.3 | 2.9.0 / 2.9.1 |
| --- | --- | --- |
| `/(?:^)bc/`, `/(?i)^bc/` | `'bc'` | `null` |
| `/\bbc/` | `'bc'` | `null` |
| `/(?<![a-z])bc/` | `'bc'` | `null` |
| `/\Abc/` | `'bc'` | `null` |
| `/^cd/m` | `'cd'` | `null` |
| `/\Bbc/` | `null` | `'bc'` |
| `/(?<=a)bc/` | `null` | `'bc'` |

Note the last two rows: the change cuts both ways, producing spurious matches as well as missed ones, and `/\b\w+/` mid-word returns a different match with a different cursor advance. No built-in parser uses any of these constructs — every in-tree pattern is either leading-`^` or purely forward-looking, which is why the test suite never noticed — but `Cursor` is public API and extensions supply their own patterns.

### Why not just revert

The in-place path is what keeps repeated `match()` calls linear. Forcing every pattern back onto the old copy-the-remainder path, on a paragraph of backtick runs whose opener can never be closed:

| n | in place | copied remainder |
| --- | --- | --- |
| 100,000 | 373 ms | 837 ms |
| 400,000 | 1,605 ms | 10,376 ms |

The gap widens with each doubling — the copy path re-copies everything left in the block on every call.

### The fix

`match()` now classifies each pattern once (memoized) as offset-safe or not:

- **Rejected outright**: `\b`, `\B`, `\A`, lookbehinds, a leading `^` combined with the `m` modifier, bracket-style or `^` delimiters, and anything that could contain a comment — the `x` modifier (trailing or via an inline flag group) or an `(?#...)` group — since a comment is zero-width yet can hide a `[` that would make the anchor scan below swallow a real anchor.
- **Anchor scan**: escape pairs and character classes — where a `^` is a literal, not an anchor — are stripped, then any `^` other than the leading one `match()` rewrites into `\G` marks the pattern offset-sensitive. Only a `^` at the original position 1 is exempt, so a `^` that becomes leading after something zero-width is stripped (such as `\G`) still counts.

Offset-sensitive patterns take the pre-2.9 copy-based path, restoring exact 2.8.3 semantics (including its partially-consumed-tab column arithmetic, which is now conditional again since the copy path no longer implies a tab). The classifier is conservative by construction: any pattern it cannot analyze — unusual delimiter, stripper failure — falls back to the always-correct copy, so a misjudgment can only cost speed, never correctness.

All built-in patterns classify as offset-safe, so core parsing keeps the fast path; the classifier itself measured +0.4% on an ASCII document and +0.02% on a multibyte one, within noise. The memo is bounded (1,024 patterns, 1,024 bytes each, ~1MB worst case) so a caller interpolating variable text into its patterns cannot grow it without bound.

### Testing

- 26 new `Cursor::match()` unit cases covering every construct above plus tab-spanning matches, multibyte subjects, alternation anchors, bracket-delimited patterns, and anchors hidden behind comments — asserting both the returned match and the resulting cursor position. 19 of them fail against the current `2.9`.
- New pathological case, `Unclosable backtick opener followed by many runs`, which guards the other direction: it times out if `match()` ever goes back to copying the remainder per call. Unlike the existing `Backticks` case, its input stays linear in the number of runs, so it can be scaled far enough to make the per-run cost visible.
- `phpunit` (4,307 tests), `tests/pathological/test.php` (67 cases), and `phpcs` under PHP 7.4 all pass.

### DIFF
--- a/src/Parser/Cursor.php
+++ b/src/Parser/Cursor.php
@@ -28,6 +28,28 @@ class Cursor
      */
     private const BYTE_OFFSET_CHECKPOINT_INTERVAL = 16;
 
+    /**
+     * Matches one escape pair, or one character class including a leading "^" or "]".
+     * Removing both from a pattern leaves only its structural characters behind.
+     */
+    private const ESCAPES_AND_CHARACTER_CLASSES = '/\\\\.|\[\^?\]?(?:[^\]\\\\]|\\\\.)*\]/s';
+
+    /**
+     * Upper bound on the number of distinct patterns whose match() strategy is remembered.
+     * Far above any realistic parser registry (the built-in parsers contribute about sixteen
+     * patterns), so it exists only to bound the cache against a caller that interpolates
+     * variable text into a new pattern on every call.
+     */
+    private const MAX_MEMOIZED_PATTERNS = 1024;
+
+    /**
+     * Upper bound (in bytes) on the length of a pattern eligible for memoization. Cache keys
+     * are the pattern strings themselves, so together with MAX_MEMOIZED_PATTERNS this bounds
+     * the cache's worst-case memory at about 1MB. The longest built-in pattern is 174 bytes;
+     * an oversized pattern is still classified correctly on every call, just never cached.
+     */
+    private const MAX_MEMOIZED_PATTERN_LENGTH = 1024;
+
     /** @psalm-readonly */
     private string $line;
 
@@ -91,6 +113,13 @@ class Cursor
      * @var array<int, string>
      */
     private array $charCache = [];
+
+    /**
+     * Cached isRegexOffsetSafe() verdicts, keyed by pattern.
+     *
+     * @var array<string, bool>
+     */
+    private static array $regexOffsetSafetyCache = [];
 
     /**
      * @param string $line The line being parsed (ASCII or UTF-8)
@@ -548,9 +577,10 @@ class Cursor
     {
         // When a tab has been partially consumed the remainder is reconstructed with the
         // leftover tab expanded into spaces, so matching must run against that reconstructed
-        // string rather than the raw line. This is rare; use the copy-based path to preserve
-        // the exact column arithmetic.
-        if ($this->partiallyConsumedTab) {
+        // string rather than the raw line. Patterns that inspect what precedes the match start
+        // must also see a subject that begins at the cursor. Both are rare; use the copy-based
+        // path for them.
+        if ($this->partiallyConsumedTab || ! self::isRegexOffsetSafe($regex)) {
             return $this->matchViaRemainder($regex);
         }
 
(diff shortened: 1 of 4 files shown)


src/Parser/Cursor.php
```diff
@@ -589,9 +619,120 @@ public function match(string $regex): ?string
...
+        if ($regex[1] === '^' && \strpos($modifiers, 'm') !== false) {
+            return false;
+        }
+
+        if (
+            \strpos($regex, '\\b') !== false
+            || \strpos($regex, '\\B') !== false
+            || \strpos($regex, '\\A') !== false
+            || \strpos($regex, '(?<=') !== false
+            || \strpos($regex, '(?<!') !== false
```
GrahamCampbell, line 689: I think `\X` needs to use the remainder path too. PCRE2 normally processes graphemes forwards, but its Regional Indicator handling scans back to the actual start of the subject to determine pairing. Given three indicators and a cursor at position 1, `/^\X/u` matches the final two against the old remainder, but only one against the whole subject at an offset.

The fallback could be extended with:

```php
|| \strpos($regex, '\\X') !== false
```

A focused regression case would be:

```php
'regional indicator grapheme' => ["\u{1F1E6}\u{1F1E7}\u{1F1E8}", '/^\X/u', 1, 3, "\u{1F1E7}\u{1F1E8}"],
```

src/Parser/Cursor.php
```diff
@@ -589,9 +619,120 @@ public function match(string $regex): ?string
...
+            return false;
+        }
+
+        // Strip escape pairs and character classes, where a "^" is a literal rather than an
+        // anchor, then look for an anchor other than the leading one match() rewrites into "\G".
+        // A "^" that only becomes leading once something zero-width is stripped from in front of
+        // it (such as "\G") is still a real anchor, so only a "^" at the original position 1 is
+        // exempt from the scan. A null from the stripper means the pattern could not be analyzed,
+        // which must classify it as offset-sensitive rather than safe.
+        $anchorsOnly = \preg_replace(self::ESCAPES_AND_CHARACTER_CLASSES, '', $regex);
```
GrahamCampbell, line 707: PCRE2 treats `[[:<:]]` and `[[:>:]]` as word-boundary assertions, but this pass treats them as ordinary character classes and can therefore mark them as offset-safe. For example, `/[[:<:]]bc/` at position 1 in `abcdef` should match `bc` when the cursor is treated as the start of the subject, but it fails when the preceding `a` remains visible.

The conservative fix could simply reject both spellings:

```php
|| \strpos($regex, '[[:<:]]') !== false
|| \strpos($regex, '[[:>:]]') !== false
```

These cases would cover both directions:

```php
'POSIX start-of-word boundary' => ['abcdef', '/[[:<:]]bc/', 1, 3, 'bc'],
'POSIX end-of-word boundary'   => ['a bc', '/[[:>:]] bc/', 1, 1, null],
```

colinodell:
This doesn't feel like the right approach, closing and trying something else.

PR #1146, opened 2026-08-10
Fix 2.9.0 regression: restore Cursor::match() remainder semantics

### LINKED ISSUE #1145: Cursor::match() makes text before the cursor visible to patterns (regression in 2.9.0)
### A regression was introduced in 2.9.0

As part of that release's performance work, `Cursor::match()` stopped matching against a copy of the remainder and began matching against the whole persistent line at the cursor's byte offset (rewriting a leading `^` to `\G` to compensate). This avoided an O(n²) copy-per-call cost in scanning loops — but it also silently changed what certain regex constructs mean, because PCRE keeps the text *before* a start offset visible to the pattern.

No built-in parser is affected — every in-tree pattern happens to be insensitive to the change — but `Cursor::match()` is public API, and extensions supply their own patterns. For example, at position 1 of `"abcdef"`, `/\bbc/` matched `'bc'` in 2.8.3 but returns `null` in 2.9.x, while `/(?<=a)bc/` did *not* match in 2.8.3 but now does. Missed matches, spurious matches, and shifted matches are all possible.

### Two contracts

There are two coherent ways a cursor-based `match()` can treat the text before the cursor:

- **Contract A — "the subject starts at the cursor."** The pattern sees a string that begins at the cursor; earlier text does not exist. `^` and `\A` anchor at the cursor; `\b` and lookbehinds see a subject boundary there. This was the behavior (and documented mental model) of every release through 2.8.x. It's simple and matches the "consume forward from here" style of parser regexes, but it *fabricates* boundaries: it answers context questions about a copy rather than about the document.
- **Contract B — "a position within the line."** Matching starts at the cursor but the line is the subject: `\G` anchors at the cursor, `^` means start of line, and `\b`/lookbehinds see the real preceding text. This is PCRE's native offset behavior and what other ecosystems deliberately chose for lexing (Python's `re.match(pos=…)`, JS sticky `/y`, Java's `Matcher.region`) — and it's genuinely useful for Markdown, where left-context questions (flanking-style rules) come up constantly.

**What we have today (2.9.0/2.9.1) is neither** — an incoherent hybrid: the leading `^` is rewritten to behave like contract A while everything else behaves like contract B. That blend matches no documentation and no user's mental model, and it's where the regression lives.

**What we really want:** `match()` restored to honest contract A (it's the documented, depended-upon behavior of a stable major version), and contract B offered *deliberately* later as a documented, opt-in API — at which point extensions get both the fast path and new expressive power (real left-context assertions) with engine-native semantics that never require the library to guess.

### The idea we tried and didn't like

#1144 kept the hybrid and tried to make it safe by classifying each pattern at runtime: offset-safe patterns match in place, everything else falls back to a remainder copy. Review kept finding classifier holes — alternative lookbehind spellings (`(*plb:…)`, `(?<*…)`), comment contexts that disguise constructs (`(?#…)`, `/x`-mode comments), `\Q…\E` quoting — and each hole is a silent wrong parse, not a slowdown. Even hardened into an allow-list, the design still interposes a guess about someone else's regex between user input and correctness. Guessing is the flaw; no classifier variant removes it.

### The approach we're pursuing instead

Move the decision to a trust boundary instead of an inference:

- **2.9.2:** `Cursor::match()` returns to the exact 2.8.3 contract-A implementation — correct for every pattern, no classification anywhere. The in-place fast path moves to a new `@internal` method used only by the core call sites, whose fixed set of patterns has been manually audited as offset-safe. Core scanning loops keep their linear performance (guarded by a new pathological test), and the 2.9.0 security fixes are unaffected — the multibyte DoS fix (GHSA-2q4p-g7hv-5rgv) lives in the byte-offset checkpoint map, which the remainder path also uses. Third-party `match()` calls return to the pre-2.9 cost profile, the status quo of every release before 2.9.0.
- **2.10.0:** introduce a public, documented contract-B method (e.g. `matchAt()`): PCRE offset semantics verbatim. Since `\G` means "at the cursor" under *both* contracts, patterns migrated to `\G` behave identically everywhere, making adoption incremental and safe.

colinodell:
Closes #1145. Supersedes #1144.

### What this does

**Fixes a regression introduced in 2.9.0** by restoring `Cursor::match()` to the exact 2.8.3 contract: **the subject begins at the cursor.** Text before the cursor is invisible to the pattern, `^`/`\A` anchor at the cursor, and `\b`/lookbehinds see a subject boundary there — for every pattern, on every PCRE version, with no runtime classification anywhere. See #1145 for the full background on the two contracts and why the classification approach from #1144 was abandoned.

The in-place fast path from 2.9.0 doesn't go away — it moves to a new `@internal` method, `Cursor::matchInPlace()`, with **PCRE's native offset semantics, verbatim**: the whole line is the subject, matching starts at the cursor, `\G` anchors at the cursor, `^` means start of line, and lookbehinds/`\b` see the characters actually preceding the cursor. No pattern rewriting happens anywhere — the 14 core call sites switch to it with their patterns re-anchored from `^` to `\G` (byte-for-byte the same regex the old rewrite produced, so behavior is unchanged). Two call sites previously used public `^`-anchored `RegexHelper` constants, whose values can't change before 3.0. Those constants are now composed from new `@internal` *unanchored* fragment constants (`PARTIAL_LINK_TITLE_UNANCHORED`, `PARTIAL_LINK_DESTINATION_BRACES`) — the public values stay byte-identical by construction (pinned by a new BC test), and the call sites anchor the fragments with `\G` directly. In 2.10 the fragments can drop `@internal` and the anchored originals can be deprecated, aligning the whole `PARTIAL_` family on "fragments are unanchored; call sites choose the anchor." When a partially-consumed tab makes the remainder differ from the underlying line, `matchInPlace()` falls back to `match()` — where `\G` still anchors at the cursor, since it means "at the cursor" under both contracts.

This shape is deliberate: `matchInPlace()` is the engine-native "contract B" method discussed in #1145, planned to become public API in 2.10 by removing the `@internal` tag. Nothing about it requires callers to certify anything — it does exactly what PCRE documents. (It keeps the `matchInPlace` name rather than `matchAt` because `RegexHelper::matchAt()` already exists with detached-substring semantics — reusing that name for the opposite contract would invite confusion.)

### Security posture

- The multibyte quadratic-parsing fix from 2.9.0 (GHSA-2q4p-g7hv-5rgv) is unaffected: it lives in the byte-offset checkpoint map, which the restored remainder path also uses — remainder construction remains cheaper than it ever was in 2.8.x.
- The only two `match()`-in-a-loop sites in the tree (the backtick closer scan and the attribute list parser) are core call sites now on `matchInPlace()`, so pathological inputs keep their linear cost. A new pathological case (`Unclosable backtick opener followed by many runs`) times out if either ever regresses to copying the remainder per call.
- Third-party `match()` calls return to the pre-2.9 cost profile — one remainder copy per call, the status quo of every release before 2.9.0.

### Testing

- 26 new `match()` unit cases pinning contract-A semantics for every construct family the 2.9.0 regression affected — lookbehinds, `\b`/`\B`, `\A`, mid-pattern `^` (alternations, groups, inline flags), leading `^` with `/m`, plus the scanner-hostile shapes found during review of #1144 (`(?#…)` and `/x`-mode comments, `\Q…\E` quoting) — asserting both the returned match and the resulting cursor position.
- New `matchInPlace()` tests: a parity table proving `\G`-anchored and forward-looking patterns behave identically through `match()` and `matchInPlace()`; an engine-native semantics table pinning contract B (`^`/`\A` anchor to the line, `^` under `/m` matches later line starts, lookbehinds and `\b` see real preceding text); and the partially-consumed-tab fallback.
- A BC-pin test asserting the recomposed public `RegexHelper` constants keep their exact historical values.
- `phpunit` (4,327 tests), `tests/pathological/test.php` (67 cases), PHPStan/Psalm at baseline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Extension/Attributes/Util/AttributesHelper.php
+++ b/src/Extension/Attributes/Util/AttributesHelper.php
@@ -24,7 +24,7 @@ use League\CommonMark\Util\RegexHelper;
 final class AttributesHelper
 {
     private const SINGLE_ATTRIBUTE = '\s*([.]-?[_a-z][^\s.}]*|[#][^\s}]+|' . RegexHelper::PARTIAL_ATTRIBUTENAME . RegexHelper::PARTIAL_ATTRIBUTEVALUESPEC . ')\s*';
-    private const ATTRIBUTE_LIST   = '/^{:?(' . self::SINGLE_ATTRIBUTE . ')+}/i';
+    private const ATTRIBUTE_LIST   = '/\G{:?(' . self::SINGLE_ATTRIBUTE . ')+}/i';
 
     /**
      * PCRE's `\s` matches the form feed that PHP's default trim charlist omits, so the
@@ -53,7 +53,7 @@ final class AttributesHelper
         // matching individual attributes since they won't need to look ahead for the closing '}'
         // while dealing with the fact that attributes can technically contain curly braces.
         // So we'll just match the start and end braces up front.
-        $attributeExpression = $cursor->match(self::ATTRIBUTE_LIST);
+        $attributeExpression = $cursor->matchInPlace(self::ATTRIBUTE_LIST);
         if ($attributeExpression === null) {
             $cursor->restoreState($state);
 
@@ -66,7 +66,7 @@ final class AttributesHelper
 
         /** @var array<string, mixed> $attributes */
         $attributes = [];
-        while ($attribute = \trim((string) $attributeCursor->match('/^' . self::SINGLE_ATTRIBUTE . '/i'), self::WHITESPACE)) {
+        while ($attribute = \trim((string) $attributeCursor->matchInPlace('/\G' . self::SINGLE_ATTRIBUTE . '/i'), self::WHITESPACE)) {
             if ($attribute[0] === '#') {
                 $attributes['id'] = \substr($attribute, 1);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Updates should follow the [Keep a CHANGELOG](https://keepachangelog.com/) princi
 - Optimized inline link destination parsing to scan the line in place, so its cost follows the length of the destination rather than the length of everything left in the block
 
 ### Fixed
+- Fixed a regression introduced in 2.9.0 where `Cursor::match()` treated text before the cursor as part of the match subject (#1145). Patterns were matched against the whole line at an offset, which silently changed the meaning of `\b`, `\B`, `\A`, lookbehinds, a `^` anywhere other than the very start of the pattern, and a leading `^` combined with the `m` modifier. `match()` once again matches against the remainder, exactly as it did in 2.8; the core parsers keep the optimized in-place matching via a new internal method with PCRE's native offset semantics, anchoring their patterns at the cursor with `\G`
 - Fixed heading permalinks rendered with `aria-hidden="true"` remaining in the keyboard tab order; they are now also given `tabindex="-1"`, as a focusable element removed from the accessibility tree has no accessible name to announce when focused (WCAG 4.1.2)
 - Fixed cloning a node breaking the link from the original node's children back to their parent, silently corrupting the document that node belonged to; detaching or inserting around those children afterwards could drop nodes from the tree
 - Fixed cloned nodes sharing their `data` with the node they were cloned from, so that setting an attribute on either one also set it on the other
--- a/src/Extension/CommonMark/Parser/Block/FencedCodeStartParser.php
+++ b/src/Extension/CommonMark/Parser/Block/FencedCodeStartParser.php
@@ -27,7 +27,7 @@ final class FencedCodeStartParser implements BlockStartParserInterface
         }
 
         $indent = $cursor->getIndent();
-        $fence  = $cursor->match('/^[ \t]*(?:`{3,}+(?!.*`)|~{3,})/');
+        $fence  = $cursor->matchInPlace('/\G[ \t]*(?:`{3,}+(?!.*`)|~{3,})/');
         if ($fence === null) {
             return BlockStart::none();
         }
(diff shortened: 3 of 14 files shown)


src/Extension/Attributes/Util/AttributesHelper.php
```diff
@@ -66,7 +66,7 @@ public static function parseAttributes(Cursor $cursor): array
 
         /** @var array<string, mixed> $attributes */
         $attributes = [];
-        while ($attribute = \trim((string) $attributeCursor->match('/^' . self::SINGLE_ATTRIBUTE . '/i'), self::WHITESPACE)) {
+        while ($attribute = \trim((string) $attributeCursor->matchInPlace('/\G' . self::SINGLE_ATTRIBUTE . '/i'), self::WHITESPACE)) {
```
GrahamCampbell, line 69: We should add a pathological case with many attributes inside one list. The existing attribute cases use many separate one-attribute lists, so they protect the outer parsing path but would not catch this inner loop reverting to `match()` and repeatedly copying a shrinking remainder. Something along these lines should exercise the behavior directly:

```php
'Many attributes in one list' => [
    'extension' => 'attributes',
    'sizes' => [1_000, 4_000, 16_000],
    'input' => static fn($n) => '{' . \str_repeat('#a ', $n) . '}',
    'expected' => static fn($n) => '<p id="a"></p>',
],
```

colinodell: Thanks for the suggestion! Added in 4cfc1148, with two tweaks: a trailing line so the attributes have a target, and sizes kept under ~2,400. Beyond that the outer match exhausts `pcre.backtrack_limit` and the list falls back to a literal paragraph. That same limit caps the inner loop's worst case at ~110ms even if it regressed to `match()`, so this is output coverage rather than a perf tripwire. I also added an "Oversized attribute list" case asserting the fast fallback, which is what actually bounds this loop.

Issue #1023, opened 2024-05-05
addRenderer and addExtension inconsistency


Arcesilas:
### Version(s) affected

2.4.2

### Description

When adding an additional custom renderer, it is not executed if no priority is set when added via an extension.

Renderer added with: `$environment->addRenderer($class, $renderer)` will be actually run on the appropriate node class, whereas renderer added via the register method of an extension will not be run at all.

Dump of `$environment->getRenderersForClass($class)` shows that the order of the renderers for the class if not the same when added via `addRenderer()` or via an extension.

When renderer is added via an extension, if it is added with a priority > 0, it will take precedence on the default renderer and be executed. It's ok, but adding a renderer via both methods should produce the same result, especially as renderer is added with the same priority in both cases by default.

If this behavior is not bogus, then maybe it should be explained in the documentation: I've spent about 2 hours trying to figure out why my extension was not working while in my previous tests with addRenderer the renderer was working fine...

### How to reproduce

Create a custom renderer. For example:
```
class YamlCodeRenderer implements NodeRendererInterface
{
    public function render(Node $node, ChildNodeRendererInterface $childRenderer)
    {
        if (str_starts_with($node->getInfo(), 'yaml')) {
            return "We should do something with YAML code block";
        }
    }
}
```
Register the renderer:
```
$cmark = new \League\CommonMark\CommonMarkConverter();
$cmark->getEnvironment()->addRenderer(FencedCode::class, new YamlCodeRenderer());
echo $cmark->convert("```yaml\ntest\n```");
```
Output:
```
We should do something with YAML code block
```

Now, register the renderer via an extension:
```
class MyExtension implements ExtensionInterface
{
    public function register(EnvironmentBuilderInterface $environment): void
    {
        $environment->addRenderer(
            FencedCode::class,
            new YamlCodeRenderer()
        );
    }
}
```
and:
```
$cmark = new \League\CommonMark\CommonMarkConverter();
$cmark->getEnvironment()->addExtension(new MyExtension());
echo $cmark->convert("```yaml\ntest\n```");
```
Output:
```
test
```

### Additional context

Dump of `$environment->getRenderersForClass()` when added via `$environment->addRenderer()`:

```
class League\CommonMark\Util\PrioritizedList#40 (2) {
  private array $list =>
  array(1) {
    [0] =>
    array(2) {
      [0] =>
      class YamlCodeRenderer#39 (0) {
      }
      [1] =>
[...]

PR #1152, opened 2026-09-08
docs: clarify priority ordering behavior

asllanmaciel:
Clarifies how component priorities should be used when customizing the environment.

The documentation now explains that:
- higher priorities are attempted first;
- block start parsers, inline parsers, and renderers should not rely on registration order when priorities tie;
- explicit, different priorities should be used whenever execution order matters;
- event listeners retain their documented same-priority registration order;
- the rendering guide explicitly calls out the renderer tie case that caused the confusion in #1023.

This is documentation-only and does not change runtime behavior.

Verification:
- `git diff --check` — pass
- documentation consistency checks for the `#priority` links, tie guidance, and event-listener exception — pass
- branch is 1 commit ahead / 0 behind `2.10`

AI assistance was used to help review the relevant priority semantics and prepare the documentation patch; the final diff and source behavior were reviewed directly.

Refs #1023.

### DIFF
--- a/docs/2.x/customization/environment.md
+++ b/docs/2.x/customization/environment.md
@@ -66,7 +66,7 @@ $environment->addExtension(new FootnoteExtension());
 public function addBlockStartParser(BlockStartParserInterface $parser, int $priority = 0);
 ```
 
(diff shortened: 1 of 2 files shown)


colinodell:
Thank you for the contribution! Overall, these changes look great and will be very helpful to users.

I'm adding a few minor tweaks, but nothing that significantly changes what you've added.

PR #530, opened 2020-08-09
Fix hard-coded footnote ID config

bakerkretzmar:
This PR fixes #524 by removing hard-coded `fn:` and `fnref:` and ensuring that the `footnote/footnote_id_prefix` and `footnote/footnote_ref_prefix` configuration options are always used.

It updates the `AnonymousFootnotesListener`, `GatherFootnotesListener`, `AnonymousFootnoteRefParser`, and `FootnoteRefParser` classes to all be configuration-aware.

I added an integration test in the style of [`Strikethrough/IntegrationTest`](https://github.com/thephpleague/commonmark/blob/latest/tests/unit/Extension/Strikethrough/IntegrationTest.php) to make sure this works, I tried to add unit tests but I couldn't figure out how to render both a footnote _and_ its reference. I'm happy to add more tests if someone can point me in the right direction.

Issue #524, opened 2020-07-20
Footnote ID prefix and footnote ref ID prefix config options sometimes ignored


bakerkretzmar:
**Version(s) affected**: 1.5.x

**Description**  

The `footnote/ref_id_prefix` and `footnote/footnote_id_prefix` configuration options are only used to generate element `id` attributes, but are ignored for the `href` attributes corresponding to those IDs, breaking all footnote reference links and backlinks if either config option is customized.

The default values for these configuration options, `fnref:` and `fn:`, are hard-coded in 5 places:

https://github.com/thephpleague/commonmark/blob/d9fdf5f5ea370c988375e54c85a43b2668a048e6/src/Extension/Footnote/Event/GatherFootnotesListener.php#L52

https://github.com/thephpleague/commonmark/blob/d9fdf5f5ea370c988375e54c85a43b2668a048e6/src/Extension/Footnote/Event/GatherFootnotesListener.php#L95

https://github.com/thephpleague/commonmark/blob/d9fdf5f5ea370c988375e54c85a43b2668a048e6/src/Extension/Footnote/Event/AnonymousFootnotesListener.php#L46

https://github.com/thephpleague/commonmark/blob/d9fdf5f5ea370c988375e54c85a43b2668a048e6/src/Extension/Footnote/Parser/AnonymousFootnoteRefParser.php#L80

https://github.com/thephpleague/commonmark/blob/d9fdf5f5ea370c988375e54c85a43b2668a048e6/src/Extension/Footnote/Parser/FootnoteRefParser.php#L59

**How to reproduce**  

Config:

```php
    'footnote' => [
        'ref_id_prefix' => 'customfnref:',
        'footnote_id_prefix' => 'customfn:',
    ],
```

Markdown:

```md
Here[^note1]

[^note1]: There
```

Expected HTML:

```html
<p>Here<sup id="customfnref:note1"><a class="custom-ref" href="#customfn:note1" role="doc-noteref">1</a></sup></p>
<div class="custom-notes" role="doc-endnotes">
    <ol>
        <li class="custom-footnote" id="customfn:note1" role="doc-endnote">
            <p>There&nbsp;<a class="custom-backref" rev="footnote" href="#customfnref:note1" role="doc-backlink">↩</a></p>
        </li>
    </ol>
</div>
```

Actual HTML:

```html
<!-- Correct id, wrong href -->
[...]

PR #1153, opened 2026-09-15
Keep the configured footnote id prefixes out of the href lowercasing

youdie006:
With `footnote_id_prefix => 'Fn:'` and `ref_id_prefix => 'Fnref:'`, every footnote link and backlink on the page is a dead anchor:

```html
<sup id="Fnref:a"><a class="footnote-ref" href="#fn:a">1</a></sup>
<li class="footnote" id="Fn:a"><p>note <a class="footnote-backref" href="#fnref:a">&#8617;</a></p></li>
```

Both renderers run `mb_strtolower` over the whole destination, prefix included, while the ids eight lines away emit the configured prefix verbatim. `FootnoteRefRenderer` does both, in the same method.

This is the breakage #524 was filed for. #530 fixed the four producers of the destination string and left both renderers, and the test row it added uses all-lowercase prefixes (`customfn:` / `customfnref:`), so the remaining path is invisible to it. Hard-coding the prefixes was proposed in that thread and rejected by @markhalliwell, so the configurability is adjudicated intent, not my reading.

The prefix is sliced off the destination rather than rebuilt from the label, because `NumberFootnotesListener` rewrites the label to `"<label>__2"` for repeat references while leaving the destination alone; a label-based href would emit `#fn:note1__2` and break that case.

<details>
<summary>Verification</summary>

PHP 8.5, `vendor/bin/phpunit --no-coverage`.

| row | `FootnoteRefRenderer.php` md5 | result |
|---|---|---|
| pristine, pristine tests | `3da7bdc60efb88732838eb04472e0575` | OK, 4397 tests |
| pristine + the two new rows | `3da7bdc6…` | **1 failure**: expected `href="#Fn:note1"`, got `href="#fn:note1"` |
| this change | `8d5dd878dcd42f38d91aeaf17a62eef9` | **OK, 4399 tests**, same 6 deprecations / 91 notices / 6 skipped as pristine |

Two rows were added and they kill different mutants, so neither is redundant:

| mutant | which row dies |
|---|---|
| plain revert of `src/` | the uppercase-**prefix** row: `#Fn:note1` expected, `#fn:note1` emitted |
| drop the lowercasing entirely | the uppercase-**label** row: `id="fn:note1"` against `href="#fn:Note1"` |

The second row is a guard: it passes on master, so it is not the bug, but it rejects the obvious "stop lowercasing" fix. Before adding it I grepped `tests/` for a footnote label containing an uppercase letter and found none, so that half was unguarded.

PHPStan reports the same 4 pre-existing errors with and without this change (artefacts of running PHP 8.5 rather than the pinned CI version); it adds none.

Not run: `phpcs`. `CLAUDE.md` says it must run under PHP 7.4 and no 7.4 image was available here, so the style gate is unverified. The added lines follow the surrounding style.

</details>

Disclosure: written with AI assistance (Claude). Every number above is measured.


youdie006:
I broke the 7.4 legs and have pushed a fix in `340a5b15`. Reporting what happened rather than just the patch.

`substr()` returns `false`, not `""`, when the offset is past the end of the string on PHP < 8.0. `FootnoteBackrefRendererTest` builds a `Reference` whose destination is `dest`, which is shorter than the configured prefix, so my slice handed a bool to `mb_strtolower` and raised a `TypeError` on the 7.4 legs. I had only run the suite on PHP 8.5, where the same expression yields an empty string and nothing goes wrong.

| | `substr('dest', 7)` |
|---|---|
| PHP 7.4.33 | `false` |
| PHP 8.3.33 | `''` |

The fix strips the prefix only when the destination actually starts with it, and otherwise leaves the previous behaviour untouched. That is better semantics as well as a version fix: a destination that does not carry the configured prefix no longer gets the prefix glued onto it.

**Verified on 7.4 this time, not just asserted.** Both rows run in the same `php:7.4-alpine` container with the same `composer update`:

| tree | PHP 7.4.33 |
|---|---|
| without the guard, i.e. what broke CI | `Tests: 4399, Errors: 3` |
| `340a5b15` | **`Tests: 4399`, 0 failures** |

Still 4399 tests and 0 failures on PHP 8.5 as well.

One note: the workflow runs on the new head are sitting at `action_required`, so your CI has not re-run yet. The numbers above are mine, from the two containers, not from this PR's checks.

colinodell (approved):
Thank you!

